PR #2, opened 2023-09-25
Pointer-Reversal Marking

Abhiroop:
This adds the pointer-reversal marking algorithm that does not use the C stack. It adds a `done` field to the node representation (UNIONPTR). The `done` field captures 3 states:

   done = 0 -- none of the children marked
   done = 1 -- the left children and their successors marked
   done = 2 -- all children marked

Instead of a 4-byte int field for `done`, perhaps some bits can be stolen from `ufun`? We require simply 2 bits;

The GCRED reductions are absent from the current algorithm so the algorithm can be first tested correctly without too many ifdefs. 
It seems to pass `make everytest` now. It may require more testing and code review.


augustss:
This change is too old now.  It also needs some changes to not increase the node size.

PR #7, opened 2023-11-28
Adds emscripten Makefile and CI entries

dmjio:
### Additions
- Adds two CI entries (osx + linux) for running `mhs` tests via `node` and comparing outputs
- Adds `Makefile.emscripten` in the top-level directory (invoked using `make emscripten` from `Makefile`)
- `git` ignores some stuff

### Caveats
- Doesn't bother to compile the compiler to JS or WASM
- Doesn't contain tests for WASM (could be added later w/ `wasmer`)
- Some tests didn't pass `Foreign`, `Floating`, (so just commented them for now -- failures described below)
- Doesn't use `mhseval` for any of the tests

### Test failure summary

- `Foreign`

```
TypeError: Cannot convert -5 to a BigInt
```

- `Floating`

```
2c2
< -1.0
---
> 0.9938211910561482
```

^ still looking into these

### Improvements 

- Wasn't immediately obvious how to cleanly reuse `tests/Makefile` w/ `emcc`, but that would be an improvement to avoid some code duplication.
- Windows support code be added later

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -118,3 +118,21 @@ jobs:
     - name: make nfib
       run: make CONF=unix-32 nfibtest
       shell: alpine.sh {0}
+
+  build-linux-emscripten:
+    runs-on: ubuntu-latest
+    steps:
+     - uses: actions/checkout@v3
+     - uses: mymindstorm/setup-emsdk@v13
+     - uses: actions/setup-node@v4
+     - name: run emscripten tests
+       run: make emscripten
+
+  build-macos-emscripten:
+    runs-on: macos-latest
+    steps:
+     - uses: actions/checkout@v3
+     - uses: mymindstorm/setup-emsdk@v13
+     - uses: actions/setup-node@v4
+     - name: run emscripten tests
+       run: make emscripten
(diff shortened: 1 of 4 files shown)


augustss:
There seems to be a lot of test failures.

augustss:
But maybe that's my fault.  Let me investigate.

augustss:
I've fixed the failing tests.  Please rerun the tests on your branch.

augustss:
Does emcc have 32 or 64 bit pointers?  Since this determines the size of Int and Word in mhs,

dmjio:
@augustss for the node / browser environments, seems to be 32-bit pointers.

```C
#include <stdio.h>

int main () {
  printf ("%zu\n", sizeof(void *));
  return 0;
}
```

```bash
$ emcc main.c -o main
$ node ./main 
4
```

vs. (I'm on an intel mac)

```
$ gcc main.c -o main
$ ./main 
8
```

augustss:
OK.  I've fixed some problems with 32 bits.  So the `Floating` test should work now.  Maybe `Foreign` too.  I've commented out the `FArith` test which is not portable between 32 and 64 bits.

PR #8, opened 2023-12-18
errtester.sh does not work on fresh clone, create  if not existing

Rewbert:


Rewbert:
closed as pr #9 includes this change

PR #9, opened 2023-12-18
Factor out some more platform specific stuff

Rewbert:
Factored out some platform specific stuff as I am running mhs on an STM32 microcontroller. This MR does not include any STM32 configuration, but it does some groundwork in preparation for it. These little microcontrollers come with their own printf's, malloc's, etc.

With these changes in place, most of the 'new code' can go into the configuration files.


tests/errtester.sh
augustss, line 1: Simpler: 'mkdir -p $tmp'

Rewbert: Agree

Rewbert:
Agree, that sounds more reasonable. I applied the changes.

PR #14, opened 2024-01-09
factor out realloc and factor out buffer allocation

Rewbert:
Trying to run this on a microcontroller and I am running out of stack...


augustss:
Does it run out of stack during parsing?
There parser could be changed to not use the C stack.

Rewbert:
Yes, during parsing. I can parse 170 something nodes before it has recursed far enough to exhaust my stack. I might be able to tweak my configurations and get access to some more stack, but the parser would still implement an upper bound on the sizes of programs we can run. It would be better to rewrite the parser to use less stack/not the C stack/not recurse in this manner(?).

augustss:
I will think about how to rewrite the parser.  It might be better to change the file format to be postfix rather than prefix.

PR #26, opened 2024-01-19
Serialize combinators use depth-first search

jmaessen:
I tried this using scc and it cost an extra 2s in bootstrap.  By contrast this hums along nicely and both DCEs and orders the output sensibly (postorder, which is reverse topological and no worse than any other simple cycle handling strategy for mutual recursion).


augustss:
You can probably tweak this to reverse the order of the definitions.

PR #28, opened 2024-01-21
Turn eval into evali

jmaessen:
This is based on the observation that two things can happen in eval:
* The stack is clear, in which case the value in n is a direct reference to the WHNF of the original argument
* The stack is not clear, in which case the top of stack is an AP representing the argument without indirections.

In either case there's no need to push / pop it again and go chasing after indirections.

This represents a small time savings of about .2s on bootstrap, see https://docs.google.com/spreadsheets/d/1Tw6LLwTEU0VxdWcf_ZUF1_T9KlchpNEFXKNiAarxd4I which shows data from 10 interleaved after/before runs plus some statistically bootstrapped averages with histograms showing limited overlap.

### DIFF
--- a/src/runtime/eval.c
+++ b/src/runtime/eval.c
@@ -1897,20 +1897,7 @@ mkStringU(struct ustring *str)
   return n;
 }
 
-void eval(NODEPTR n);
-
-/* Evaluate and skip indirections. */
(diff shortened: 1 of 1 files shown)


augustss:
Excellent observation.

PR #30, opened 2024-01-28
GC check during parse so we don't crash.

jmaessen:


augustss:
The GC check is intentionally missing. Instead we always GC before parsing.

augustss:
But the reason it was like that was the way the recursive parser worked.  Now there is no reason for that anymore, and maybe it's better with a GC check in the parser.

PR #32, opened 2024-01-29
Fix asan and ubsan violations, one of which was causing periodic crashes

jmaessen:
I decided to run sanitzers on this and found why I was getting periodic heap corruption – it was tmpname overflowing its buffer.  But it's ubsan I'm interested in as I'm screwing around with eval and gc a bit and trying not to confuse the optimizer.


augustss:
Thanks!  I've run a sanitizer too, and it didn't find these.

PR #34, opened 2024-02-01
Gc pointer stack

jmaessen:
Uses the pointer stack to perform GC.  This is slightly faster, but only if:
* You arrange the pushes for aps in the direction shown (otherwise it's a festival of stack overflows during bootstrap).
* Check the mark before pushing.  I have no idea why this makes any difference.
* Use a custom control flow path for push.  If you just check to_push at the top / bottom of the loop it's a wash.

I'm not sure in hindsight the switch to a switch was a great plan, but the perf looks good at least.


augustss:
Excellent!  The reduces that amount of C stack needed.

PR #41, opened 2024-04-05
Easy cross compiling part-2

yobson:
- Added the ability to define targets
- generate a default target when building
- added the ability to select target when invoking

I will probably make only small changes in part 3.


augustss:
I think this a great idea.
But I've been trying to keep the compiler quite lean, so I'm not entirely happy about the duplication of the lexer.
I think you can do the target parsing using the standard mhs lexer.

yobson:
I’ll see what I can do!

yobson:
I’ll probably not manage to take a look at this until Sunday.

but a few questions:
1. Are the test failures due to my code changes?
2. Do you have links to any literature do explain how your combinator runtime works?

augustss:
1.  The failing tests are my fault. I implemented ForeignPtr, but it crashes on 32 but platforms. 

2.  I recommend the original paper:
Turner, D.A. (1979). "A New Implementation Technique for Applicative Languages". Software: Practice and Experience. 9: 31–49.

augustss:
Actually, some of the failing tests might be your fault.  I've temporarily disabled my failing test.

yobson:
Hahaha. Lots to do on Sunday then 😂

yobson:
Okay! I've reworked it to use your lexer. I did this by exposing `lex` from Lex.hs. It seemed like the simple solution.

I have introduced a warning. My TokenMachine instance is orphaned. I can move the definition into Lex.hs if you want that warning gone :)

augustss:
As you can see, the emscripten tests now fail.
Cross compiling to WASM/JS will no doubt benefit from your changes.  You just need to fix something. :)

augustss:
I don't mind orphan instances.  They might not be great for a library, but within a self contained application I think they are fine.

yobson:
I see why the test is failing for the JS build.
It sets `CC`, but because `targets.conf` exists, it ignores `CC` and uses the default profile

augustss:
So emscripten should be changed to use the new target stuff.

yobson:
I also made it such that the `CC` variable and `MHSCONF` variable will override targets.conf, instead of the other way around. This seems far more natural to use, and easier to hack with.

PR #48, opened 2024-07-11
fix small bug in unQualString

Rewbert:
I believe I found a bug in `unQualString`. The `Ident` identifying instances, after typechecking, might have names such as `"inst$(Data.Typeable.Typeable@Control.Exception.Internal.PatternMatchFail)"`.

As an expression, `EVar -theident-`, the pretty printer will render this as an unqualified identifier. The function `unQualString` will look for `'.'` for where to split the string, but it will not take into account the parentheses. The rendered symbol will be `"PatternMatchFail)"`.

I just quickly added a fix for this, assuming that any opened parentheses will be closed at the end of the string.

I noticed this while I was poking about in the type checker, rendering the `[InstDef]` in `TModule`.

### DIFF
--- a/src/MicroHs/Ident.hs
+++ b/src/MicroHs/Ident.hs
@@ -79,15 +79,51 @@ addIdentSuffix (Ident loc i) s = Ident loc (i ++ s)
 
 unQualString :: HasCallStack =>
                 String -> String
-unQualString [] = ""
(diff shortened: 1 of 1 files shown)


augustss:
I don't like how much code this adds to unqualString.  I'll do something different.

PR #60, opened 2024-10-02
Fix Paths_MicroHs

kolmodin:
Use "echo -e" to enable interpreting the backslash escapes.


kolmodin:
Otherwise "make install" fails with:

```
       > mhs: error: "dist-mcabal/Paths_MicroHs.hs": line 1, col 29:
       >   found:    \
       >   expected: data newtype type LIdent ( UQIdent [ literal ~ ! import foreign infixl infixr infix class instance default pattern ; }
```

PR #74, opened 2024-12-16
Use a mergesort

treeowl:


src/MicroHs/List.hs
```diff
@@ -12,11 +12,30 @@ import Data.List
...
-sortLE _  [] = []
-sortLE le (x:xs) = sortLE le lt ++ (x : sortLE le ge)
-  where (ge, lt) = partition (le x) xs
+sortLE le = mergeAll . splatter
+  where
+    splatter [] = []
+    splatter [a] = [[a]]
+    splatter (a1 : a2 : as)
+      | a1 `le` a2 = [a1, a2] : splatter as
+      | otherwise  = [a2, a1] : splatter as
```
treeowl, line 23: In GHC, using this sort of paired splatter significantly improves performance, by avoiding allocation churn. I don't know if works the same way in MicroHs; if not, we can just use `splatter = map (: [])`.

augustss:
That looks a lot like my old merge-sort.  I think it has its roots there.

augustss:
The PR saves about 0.1% execution time and allocation.  Excellent!

PR #78, opened 2024-12-31
Implement `clamp` in Data.Ord

liolin:
I stumbled across the fact that the function `clamp` is missing in `Data.Ord`. If there is a reason why the function is not implemented, please let me know and simply close the PR.


augustss:
I'm not sure why it was missing.

PR #81, opened 2025-01-06
Fix empty lambda case

konsumlamm:
Compiling `\case {}` did throw an "impossible" exception before. Now `EmptyCase` is fully supported. Additionally, `ELam` now stores a `SLoc` (the location of the `\`), which is used for the error message in a lambda case. I regenerated `generated/mhs.c` using `make newmhs`, is that the intended workflow? (It actually took me longer than fixing the bug to find out that I had to do this in order for the compiler changes to be carried over.)

I added `bin/mhseval` to `runtestmhs` and `runtestemscripten`, since it's required by those, but I'm not very familiar with makefiles, so correct me if this is wrong.


augustss:
Thank you!
Yes, the Makefile is a mess.  I need to clean it up.
The target newmhs does indeed regenerate generated/mhs.c, but to get the compressed version (which is more suitable for committing) you need the (slower) target newmhsz.

PR #82, opened 2025-01-06
Implement NumericUnderscores

konsumlamm:
I modified the lexer to accept underscores in numeric literals and added some tests. While testing, I noticed that using `_` as a variable throws an "impossible" exception, so I added a proper error message (this is the only actual change in `TypeCheck.hs`, in line 1625). This time I updated `generated/mhs.c` with `make newmhsz`.


augustss:
I would like you to separate the changes that only affect whitespace from actual changes. 

On 32 bit platforms, Double is 32 bits. So many numeric values do not come out the same as in 64 bit.

augustss:
Hmm, looking at the code I realize that my original lexer is wrong.  It lexes '0x' is the number 0, but it should be two lexemes: 0 and x.

konsumlamm:
> Hmm, looking at the code I realize that my original lexer is wrong. It lexes '0x' is the number 0, but it should be two lexemes: 0 and x.

Yes, my PR also fixes that.

src/MicroHs/Lex.hs
```diff
@@ -80,10 +81,24 @@ lex loc (d:cs) | isLower_ d =
   case span isIdentChar cs of
     (ds, rs) -> tIdent loc [] (d:ds) (lex (addCol loc $ 1 + length ds) rs)
 lex loc cs@(d:_) | isUpper d = upperIdent loc loc [] cs
-lex loc ('0':x:cs) | toLower x == 'x' = hexNumber loc cs
-                   | toLower x == 'o' = octNumber loc cs
-                   | toLower x == 'b' = binNumber loc cs
-lex loc cs@(d:_) | isDigit d = number loc cs
+lex loc ('0':x:cs)
+  | toLower x == 'x' =
```
augustss, line 85: There's a lot of code duplication here.  You can make a new function alleviate that.

augustss (changes requested):

augustss:
That's looking good. Now I just want to check that this has not caused any significant slowdown. The lexer is fairly performance sensitive. I'm willing to accept a 0.5% slowdown for a feature like this. 

I measure it by doing ’make timecompile' before and after and checking the number of reductions.

PR #83, opened 2025-01-08
Add some missing functions/instances from the report

konsumlamm:
Additions:
* `Control.Monad`: add `liftM4`, `liftM5`
* `Data.Array`: implement the `Ord (Array a b)` instance (based on the `base` implementation)
* `Data.Char`: add `isControl`, `isLetter`
* `Data.Ratio`: add `Enum (Ratio a)` instance
* `System.Exit`: derive `Ord ExitCode` instance

I also added `numericEnumFrom`, `numericEnumFromThen`, `numericEnumFromTo` and `numericEnumFromThenTo` (as described in the report, but using your implementation) to `Data.Enum` and used them to implement other numeric `Enum` instances, reducing code duplication a lot. They are currently exported from `Data.Enum` (unlike in `base`, where they are exported from `GHC.Real`), but I figured that's okay, since they're useful for writing `Enum` instances and `boundedEnumFrom` and `boundedEnumFromThen` are also exported from `Data.Enum` (unlike in `base`, where they are exported from `GHC.Enum`).

### DIFF
--- a/lib/Data/Enum.hs
+++ b/lib/Data/Enum.hs
@@ -40,6 +40,25 @@ boundedEnumFromThen n1 n2
     i_n1 = fromEnum n1
     i_n2 = fromEnum n2
 
+numericEnumFrom :: (Num a) => a -> [a]
+numericEnumFrom n = n : numericEnumFrom (n + 1)
+
+numericEnumFromThen :: (Num a) => a -> a -> [a]
+numericEnumFromThen n m = from n
+  where
+    d = m - n
+    from i = i : from (i + d)
+
+numericEnumFromTo :: (Num a, Ord a) => a -> a -> [a]
+numericEnumFromTo l h = takeWhile (<= h) (numericEnumFrom l)
+
+numericEnumFromThenTo :: (Num a, Ord a) => a -> a -> a -> [a]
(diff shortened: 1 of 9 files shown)


lib/Data/Enum.hs
```diff
@@ -40,23 +40,36 @@ boundedEnumFromThen n1 n2
...
+numericEnumFrom n = n : numericEnumFrom (n + 1)
+
+numericEnumFromThen :: (Num a) => a -> a -> [a]
+numericEnumFromThen n m = from n
+  where
+    d = m - n
+    from i = i : from (i + d)
+
+numericEnumFromTo :: (Num a, Ord a) => a -> a -> [a]
+numericEnumFromTo l h = takeWhile (<= h) (numericEnumFrom l)
```
konsumlamm, line 53: The report and `base` use `h + 1/2` here, to account for rounding errors, but since there are currently no `Enum` instances for float types, that shouldn't matter, so I kept it like this.

augustss: Thank you!  I'll merge this ASAP.

augustss:
Thanks for aligning MicroHs with the report/ghc.  Keep up the good work.

PR #85, opened 2025-01-10
Add missing primitives in `translate` & remove unused primitive

konsumlamm:
Add the `ucmp` and `A.copy` primitives to the `primTable` in `MicroHs.Translate`, so that they can be used in the REPL and with `-r`.

I noticed that some of the primops are never used:
* `IO.getTimeMilli`: This has been replaced with FFI, so I decided to remove it.
* `compare`: This is an alias for `icmp`, so I kept it for now.
* `IO.C'BIND`: I have no idea what this does, so I kept it.


augustss:
Thanks!
The `IO.C'BIND` is a special combinator used when executing IO actions.  It is only generated by the runtime itself.
There's an explanation in eval.c.

PR #87, opened 2025-01-11
Optimize `Integer` a bit

konsumlamm:
Some small optimizations for `Integer`s. Each optimization is in its own commit:
* Since each digit is nonnegative, we can use `Word` instead of `Int` to represent digits. This allows us to use half of the bits (16/32) without multiplication overflow. It also has the effect that any `Int` and `Word` can be represented using at most 2 digits.
* Since `maxD` is a power of 2, we can use bit operations instead of `quot` and `rem` (see `quotMaxD`, `remMaxD`).
* `dropWhileEnd (== 0)` seems to be a bit more efficient than `reverse . dropWhile (== 0) . reverse`.
* `quotRemI` used to traverse `ys` 4 times in the case that all digits except the MSD are 0: once for `init`, `last`, `length` and `all`. `msd` does all this in a single traversal.
* ~~`_integerToIntList` and `_intListToInteger` used costly `Integer` division and multiplication. By changing the base to 2^16=32768, we can either reuse the list (on 32 bit platforms) or split/merge 2 elements (on 64 bit platforms), avoiding any expensive operations.~~

I tried to rewrite `Integer` to use arrays (using my own array type based on `IOArray`), however that implementation was quite a bit slower than the current list-based implementation. Implementing `Integer` in the runtime should make it significantly faster, be it by using an existing library (e.g. GMP, like GHC does by default) or by manually writing a C implementation (I'd be generally willing to try this), but I'm not sure if this is even desirable, since MicroHs is supposed to be minimal.

### DIFF
--- a/lib/Data/Integer.hs
+++ b/lib/Data/Integer.hs
@@ -14,6 +14,7 @@ module Data.Integer(
 import Prelude()              -- do not import Prelude
 import Primitives
 import Control.Error
+import Data.Bits
 import Data.Bool
 import Data.Char
 import Data.Enum
@@ -148,7 +149,7 @@ add' ci []       []       = if ci == zeroD then [] else [ci]
 
 -- Add 3 digits with carry
 addD :: Digit -> Digit -> Digit -> (Digit, Digit)
-addD x y z = (quot s maxD, rem s maxD)  where s = x + y + z
+addD x y z = (quotMaxD s, remMaxD s)  where s = x + y + z
 
 -- Invariant: xs >= ys, so result is always >= 0
 sub :: [Digit] -> [Digit] -> [Digit]
@@ -164,7 +165,7 @@ sub' _  []       _        = error "impossible: xs >= ys"
 subW :: Digit -> Digit -> Digit -> (Digit, Digit)
 subW b x y =
   let d = maxD + x - y - b
-  in (1 - quot d maxD, rem d maxD)
+  in (1 - quotMaxD d, remMaxD d)
 
 -- Remove trailing 0s
 trim0 :: [Digit] -> [Digit]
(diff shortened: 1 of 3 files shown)


augustss:
If you want to make an interface to GMP, I'm all for it.  I've been thinking of doing this. Then we would make an option to choose between the GMP implementation and the Haskell one. I've not decided quite what that will look like, but it's doable. 

If you do this, make it a new type, and I will see how to best slot it in. It should be possible to use ForeignPtr and finalizers to make it work with GC.

konsumlamm:
If I understand correctly, the CI failure is because I changed the way `Integer` literals are compiled (by changing `_integerToIntList` and `_intListToInteger`). How should I proceed?

augustss:
First (now that I'm at a computer rather than my phone), to follow up on using GMP.  My idea is that this should involve (almost) no changes to eval.c.  The new Integer implementation will use FFI to call GMP.  We might need some small change to eval.c to make sure this new Integer type can be serialized, since `ForeignPtr` cannot be serialized in general.

augustss:
Don't change `_integerToIntList` and `_intListToInteger`.  There's no reason to change them, they are only used by the compiler to emit an Integer constant and then (at most) once per constant when deserialized.
They are written in terms of `Integer` operations on purpose, so they can be independent of how `Integer` is implemented.
They should be moved to their own module, though.

konsumlamm:
I created a new issue for discussing how to use GMP: #88.

augustss:
Did you run the QuickCheck tests (using ghc)?
Changing integer division is scary. :)

konsumlamm:
> Did you run the QuickCheck tests (using ghc)? Changing integer division is scary. :)

I had some trouble, but I didn't find any bugs. I didn't change the actual division algorithm, I just optimized the multiple list traversals.

PR #89, opened 2025-01-12
Fix UB in shifts & add missing `Bits` instances

konsumlamm:
In C, shifting by a negative number or a number greater than or equal to the number of bits is UB. `shiftL` and `shiftR` now guard against this. Also uncomment the `Bits` instances for `Int8`, `Int16`, `Int32` and `Int64` and add `FiniteBits` instances.


augustss:
Tack!

PR #90, opened 2025-01-14
Implement `Bits Integer` instance

konsumlamm:
Implement the `Bits Integer` instance and add `bitDefault`, `testBitDefault` and `popCountDefault` to `Data.Bits` (these are copied from `base`). Also readd the optimizations reverted in https://github.com/augustss/MicroHs/commit/2ec34c1ab7655acce61166bd267290b590e20a1a (most of that didn't even use `Data.Bits` in the first place).

The `Bits Integer` instance works like in `base`, by pretending that negative numbers are stored in 2's complement (so that they conceptually have an infinite number of leading ones).

I wrote a script to generate some test cases using `QuickCheck`, which I then ran with MicroHs (I had to use `gmhs`, since `mhs` took too long).

### DIFF
--- a/lib/Data/Bits.hs
+++ b/lib/Data/Bits.hs
@@ -93,6 +93,18 @@ class Bits b => FiniteBits b where
 
       w = finiteBitSize x
 
+bitDefault :: (Bits a, Num a) => Int -> a
+bitDefault i = 1 `shiftL` i
+
+testBitDefault :: (Bits a, Num a) => a -> Int -> Bool
+testBitDefault x i = (x .&. bit i) /= 0
+
+popCountDefault :: (Bits a, Num a) => a -> Int
+popCountDefault = go 0
+  where
+    go c 0 = c
+    go c w = go (c + 1) (w .&. (w - 1)) -- clear the least significant bit
+
 _overflowError :: a
 _overflowError = error "arithmetic overflow"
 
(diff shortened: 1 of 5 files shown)


augustss:
I will only accept the using shift instead of multiplication/division if you can show me a benchmark where it makes a noticeable difference.
That change makes the MHS "binary" (i.e., the compressed combinator file) significantly bigger since it adds dependency on Data.Bits.

For a fast Integer type we need a different implementation, e.g., gmp.

augustss:
Hmmm, I now see that with this change Integer will depend on Data.Bits because of the instance. I'm not thrilled about that, but I might have to accept that. Because I do want the Bits instance for Integer.

konsumlamm:
I'll try moving the instance to `Data.Bits`.

augustss:
Ok. We probably need some Data.Integer.Internal that exports things like logical operations and shifts.

augustss:
I'll take a look tomorrow. 

Maybe I'm obsessed with compilation speed, but that's part of the reason why it compiles about 10x faster than ghc (using gmhs).

Btw, using bin/mhs is quite slow, as you've noticed, since it compiles the needed base modules every time. But if you do ’make minstall' and then use the installed mhs it's much faster since it uses the precompiled base package.

augustss:
I've decided to take your PR as is. It's worth the extra compile time.
But I will refactor the Integer code a little to make it easier to have a GMP implementation later.

PR #93, opened 2025-01-16
Integer: Optimize comparisons

konsumlamm:
Compare `Integer`s using a single traversal of the digits, rather than computing the lengths and reversing the lists first.

### DIFF
--- a/lib/Data/Integer/Internal.hs
+++ b/lib/Data/Integer/Internal.hs
@@ -3,7 +3,7 @@
 module Data.Integer.Internal(
   Integer,
   zeroI, oneI, negOneI,
-  eqI, neI, ltI, leI, gtI, geI,
+  eqI, neI, cmpI, ltI, leI, gtI, geI,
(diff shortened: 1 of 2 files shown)


lib/Data/Integer/Internal.hs
```diff
@@ -98,15 +98,14 @@ subW b x y =
...
-ltW axs ays = lxs < lys || lxs == lys && cmp (reverse axs) (reverse ays)
-  where
-    lxs = length axs
-    lys = length ays
-    cmp (x:xs) (y:ys) = x < y || x == y && cmp xs ys
-    cmp []     []     = False
-    cmp _      _      = error "ltW.cmp"
+cmpW :: [Digit] -> [Digit] -> Ordering
+cmpW (x : xs) (y : ys) =
+  case cmpW xs ys of
```
augustss, line 103: This can be written `cmpW xs ys <> compare x y`

PR #94, opened 2025-01-18
Add popcount, clz, ctz primitives

### LINKED ISSUE #92: Builtin popcount, clz, ctz
I'm interested in adding builtins for popcount, count leading zeros (clz) and count trailing zeros (ctz), since these have specialized CPU instructions on many platforms. They would be used in the `Bits`/`FiniteBits` instances for `Int`/`Word` etc. Looking at the runtime, it seems builtins are either implemented as primops or as FFI functions. Which would be preferable here?

konsumlamm:
Closes #92.

Implement primops for popcount, clz (count leading zeros) and ctz (count trailing zeros).

They use builtins for MSVC, GCC and Clang. The default implementation has `#if`s for GCC and Clang, since those are the most common compilers, to avoid having to repeat those checks for the different platforms.


src/runtime/eval.c
```diff
@@ -127,6 +127,60 @@ FFS(bits_t x)
 }
 #endif  /* !defined(FFS) */
 
+#if !defined(POPCOUNT)
+uvalue_t POPCOUNT(uvalue_t x) {
+#if defined(__GNUC__)
+  return __builtin_popcountl(x);
+#elif defined(__clang__) && __has_builtin(__builtin_popcountl)
+  return __builtin_popcountl(x);
+#else
```
konsumlamm, line 136: (This also applies to `CLZ` and `CTZ`.)

The reason that these checks aren't combined to a single one is that not all C compilers support it. Some (e.g. MSVC) don't even allow `&&` in preprocessor conditions, but to avoid that, the default implementation would have to be duplicated. The builtins take `unsigned long`s, which have the same size as a pointer on almost all platforms, the only exception I know of is Windows (but that's handled in `config-windows-64.h` and `config-mingw-64.h`).

The fallback implementations are quite naive, but they're simple and when GCC or Clang are used (which is almost always the case), they aren't used anyway.

augustss:
Looks great!  Thanks!

PR #95, opened 2025-01-21
Fix `Data.List.isInfixOf`

konsumlamm:


augustss:
Thanks!  I'm glad you're helping to improve MicroHs.

PR #96, opened 2025-01-23
Fix strictness of `foldl'` & `foldr'`

konsumlamm:
Since they used `foldl`/`foldr`, they were not actually strict. I also tried the versions used in `base` (using `foldl` for `foldr'` and `foldr` for `foldl'`), but they were worse in my micro-benchmark (I ran `print $ foldl' (+) 0 [1..10_000::Int]`/`print $ foldr' (+) 0 [1..10_000::Int]` and checked the number of reductions and memory usage), although the alternative `foldl'` actually uses less memory.


augustss:
Oops. Thanks!

PR #98, opened 2025-01-24
Export `evaluate` and `mapException` from `Control.Exception`

konsumlamm:
I don't see any reason why they're not exported.


augustss:
Thanks!

PR #102, opened 2025-01-27
Better `IsString` instance for `String`

konsumlamm:
Change to `instance (a ~ Char) => IsString [a]`, so that the string type can be inferred if it is known to be a list (e.g. in `length "hello"`). This is also in `base`, see https://hackage.haskell.org/package/ghc-internal-9.1201.0/docs/src/GHC.Internal.Data.String.html#line-63.

### DIFF
--- a/lib/Data/String.hs
+++ b/lib/Data/String.hs
@@ -14,5 +14,5 @@ import Data.List
 class IsString a where
   fromString :: String -> a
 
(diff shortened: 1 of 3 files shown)


augustss:
Yes, that's better. When I wrote the IsString instance the type checker didn't support ~ yet.

PR #106, opened 2025-01-29
Switch to git submodule for cpphs

Superstar64:
Use git submodule instead of git clone for fetching cpphs. This is better for reproducability as cpphs may change in a MicroHs incompatible way in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,4 +24,3 @@ targets.conf
 timing
 mdist/*
 dist-mcabal/
-cpphssrc/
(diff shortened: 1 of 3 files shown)


augustss:
It doesn't work

```
augustss@Lennarts-MBP MicroHs % make bootstrapcpphs
git submodule update --init --depth 1 cpphssrc/malcolm-wallace-universe
error: pathspec 'cpphssrc/malcolm-wallace-universe' did not match any file(s) known to git
make: *** [cpphssrc/malcolm-wallace-universe/.git] Error 1
```

Superstar64:
This is super embarrassing. I never actually added the submodule to the pull request. I must have done a `git reset` writing while this and git erased the module.

PR #107, opened 2025-01-29
Properly bootstrap MicroCabal

Superstar64:
Scrap `generated/mcabal.c` and instead include MicroCabal via a git submodule and use MicroHs to build it.


augustss:
I'm takeing the submodule idea, but I'm keeping how the rest works.  For now.

PR #111, opened 2025-01-29
Fix `Enum` deriving

konsumlamm:
Fix `Enum` deriving, specifically `enumFrom` and `enumFromThen`, to respect the upper/lower bounds.


src/MicroHs/Deriving.hs
```diff
@@ -251,20 +251,36 @@ derBounded _ (c, _) _ e = cannotDerive "Bounded" c e
...
+      iEnumFromThen = mkIdentSLoc loc "enumFromThen"
+      iEnumFromTo = mkIdentSLoc loc "enumFromTo"
+      iEnumFromThenTo = mkIdentSLoc loc "enumFromThenTo"
       fromEqns = zipWith mkFrom cs [0..]
       toEqns   = zipWith mkTo   cs [0..]
-      inst = Instance hdr [Fcn iFromEnum fromEqns, Fcn iToEnum toEqns]
+      enumFromEqn =
+        -- enumFrom x = enumFromTo x (last cs)
+        let x = EVar (mkIdentSLoc loc "x")
+        in eEqn [x] (eAppI2 iEnumFromTo x eLastCon)
```
konsumlamm, line 276: Does `enumFromTo` have to be qualified with `B@` here?

augustss: Yes. Any global identifiers on the RHS should be qualified. (It's unlikely to go wrong here, but it could.)

PR #115, opened 2025-02-02
Remove duplicate `NFData (Ratio a)` instance

konsumlamm:
The CI currently fails because there are two `NFData (Ratio a)` instances.


augustss:
I forgot to push.

PR #116, opened 2025-02-02
Fix/improve `Enum` instances

konsumlamm:
Change the following `Enum` instances:
* `Int(N)`, `Word(N)`
  - `succ maxBound` and `pred minBound` now error (as opposed to wrapping around), as demanded by the report
  - fix `enumFrom` and `enumFromThen` not stopping at the bounds, producing infinite lists
  - fix `enumFromTo` and `enumFromThenTo` for bounds near `minBound`/`maxBound`
* `Bool`, `Char`, `Ordering`
  - use `boundedEnumFrom` and `boundedEnumFromThen` to avoid `toEnum` errors when reaching the bounds
* `Natural`
  - use `Enum Integer` instance instead of going through `toEnum`, which breaks for `Natural`s greater than `maxBound :: Int`
  - stop at 0 when going down in `enumFromThen`
* `Fixed a`
  - use `coerce` instead of `fmap MkFixed`


augustss:
Thanks!

PR #119, opened 2025-02-03
Fix double free in `bs++` & small bytestring changes

konsumlamm:
* The optimization in `bsappend` to return one bytestring if the size of the other is zero causes a double free, since the resulting bytestring gets a new finalizer, so that there are two finalizers for the same data. The implementation of `Data.ByteString.copy` also assumes that `bs++` always allocates a new bytestring.
* Remove unused `bs+++` primitive. It is currently not implemented and likely wouldn't improve performance enough to be worth it.
* Add `instance NFData ByteString`. The default implementation using `seq` works, since `ByteString` is a primitive type which is strict in its data.


augustss:
Sorry about the double free.  I'm surprised that nasty little bug has not reared its head before.

augustss:
And a big thanks for improving bytestring!

PR #121, opened 2025-02-04
Uncomment `PrintfArg Double` instance

konsumlamm:
Make it possible to use `Double` with  `printf`. Is there any reason why it was commented out?


augustss:
Probably a remnant from before there was full support for Double.

PR #126, opened 2025-02-07
`readb` & `writeb` for more `BFILE`s

konsumlamm:
* Add `readb` for block reads
* Implement `readb` & `writeb` for more `BFILE`s (all except `BFILE_rle`, since I'm too lazy and it isn't used anyway)
* Reuse `struct bfbuffer` in `BFILE_buffer`

### DIFF
--- a/src/runtime/bfile.c
+++ b/src/runtime/bfile.c
@@ -40,6 +40,14 @@ bfbuffer_init(struct bfbuffer *bf, size_t size)
     ERR("bfbuffer_init");
 }
 
+int
+bfbuffer_get(struct bfbuffer *bf)
+{
+  if (bf->pos >= bf->size)
+    return -1;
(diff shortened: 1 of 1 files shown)


src/runtime/bfile.c
```diff
@@ -52,6 +60,37 @@ bfbuffer_snoc(struct bfbuffer *bf, int byte)
   bf->buf[bf->pos++] = byte;
 }
 
+size_t
+bfbuffer_read(struct bfbuffer *bf, uint8_t *buf, size_t size)
+{
+  if (bf->pos + size > bf->size) {
```
augustss, line 66: Can't you just say `if (bf->pos + size > bf->size) { size = bf->size - bf->pos; }` ?

konsumlamm: Yes, that would also work.

Issue #112, opened 2025-02-01
More efficient IO


konsumlamm:
IO currently does everything byte by byte. This could be significantly sped up by reading/writing multiple bytes at once, especially for bytestrings (I made a small experiment of writing a 1 MiB bytestring to a file using `fwrite` and it was ~9x faster than the current implementation). However I'm not sure how well this would work with the `BFILE` abstraction.

augustss:
BFILE needs some updates to support block writes. Writing speed has not bothered me.

augustss:
I've added some support for `writeb()` which will write multiple bytes.
Totally untested.

konsumlamm:
Ok, after spending half a day tracking down a double free, I think I have a working prototype for block reads/writes that works with bytestrings. I'll try to make a PR once all the other changes I have lying around are done.

Your `writeb` looks like a good start (at least it's very similar to what I've come up with).

PR #129, opened 2025-02-10
ByteString: More efficient IO read/write

konsumlamm:
Implement `hPut`, `hGet`, `hGetContents` using block IO and change `readFile`, `writeFile`, `appendFile` to use them.

Implements https://github.com/augustss/MicroHs/issues/112.

The code currently compiles, but it doesn't work, since `primBS2FPtr`, `primFPtr2BS` & `primFPtrLen2BS` aren't implemented correctly (`_primitive "I"` doesn't work, since while `ByteString` and `ForeignPtr` have essentially the same representation, they have different tags). I'm not sure what the best way forward is here. Ideally there would be conversion between bytestrings and forptrs without having to copy the payload, but I'm not sure that's possible. Another idea is to add `mallocByteString` and `withByteString` (so that we don't have to go through `ForeignPtr`), but that would mean code duplication.


augustss:
I think the right thing is to use the same tag for both. The tags only differ because of serialization. But that's better done with some subfield of a foreign pointer.  I think it's a straightforward change. I'll do it later today.

augustss:
I've pushed a change that gets rid of T_BSTR.

I'm on vacation in New Zealand, so responses are a bit slow. :)

konsumlamm:
Unfortunately, I think this just moved the problem. It's still not possible to simply convert between a foreign pointer and a bytestring using `I`, since the latter needs the `fptype` to be set to `FP_BSTR`. Looking at `addForPtr`, I think something similar should work here, but since the foreign pointer and the bytestring need to share the finalizer, the tag can't be stored there. I think it would've been possible before, but maybe I'm missing something.

augustss:
It can probably be fixed by just removing the sanity checks. The fptype doesn't matter, except for serialization.

konsumlamm:
> It can probably be fixed by just removing the sanity checks. The fptype doesn't matter, except for serialization.

Ok, I'll try that.

konsumlamm:
I added two new primitives, `fp2bs` & `bs2fp`. `fp2bs` takes a foreign pointer and an int and returns the pointer as a bytestring (setting the `fptype` and the `size`), while `bs2fp` simply returns the bytestring forptr unchanged. I'm not entirely sure if this is sound, but at least I haven't encountered any segfault yet.

augustss:
Just setting the type to bytestring isn't sound unless the pointer was originally a bytestring. A bytestring has a size that is used in serialization. The serialization code needs a sanity check for NOSIZE. 

I'll look at it some more tonight.

konsumlamm:
> Just setting the type to bytestring isn't sound unless the pointer was originally a bytestring. A bytestring has a size that is used in serialization. The serialization code needs a sanity check for NOSIZE.

`fp2bs` also sets the size.

augustss:
Having `fp2bs` and `bs2fp` is a good idea.

PR #141, opened 2025-02-22
Fix #140

### LINKED ISSUE #140: `Char` sometimes incorrectly needs to be in scope
While trying to build `mtl`, the following error occurred:

```
mhs: error: "./Control/Monad/Error/Class.hs": line 80, col 5: undefined type: Char
```

This is caused by the following snippet of code:

```haskell
import Prelude (Either (Left, Right), Maybe (Nothing), either, flip, (.), IO, pure, (<$>), (>>=))

class (Monad m) => MonadError e m | m -> e where
    throwError :: e -> m a
```

Importing `Char` from `Prelude` works around the bug, but isn't necessary when using GHC.

konsumlamm:
Fix #140.

Add `Char` to `Mhs.Builtin` and use `mkBuiltin` instead of `mkIdentSLoc`.

Also fix a similar bug for negative literals, adding `negate` to `Mhs.Builtin`.


augustss:
You fixed it before I woke up. :)

PR #144, opened 2025-02-22
Optimize parser

konsumlamm:
Optimize the parser by making it deterministic (i.e. allow at most one success). This is done in several steps:
- Step 1 (cd2ad121950a2cbf1ac9bc5299333cd5388f5fa8): Replace all uses of `<|>` (including transitive uses in `many`, `some`, etc.) by `<|<`, preserving the semantics by possibly changing the order in which the parsers are combined.
- Step 2 (ddad212cd930202f7a15f2af420153674593f0ee): Change the `Prsr` definition to allow at most one success. This removes the possibility of "Ambiguous" errors.
- Step 3 (58dedada0a40b26925e6e21cdca7db3b2eb861d9): Change all uses of `<|<` to `<|>`, which now has the same behaviour.

Overall, these changes make compiling mhs use 2.6% less memory and run 2.1% faster (measured using `bin/mhs -ipaths -imhs -isrc MicroHs.Main +RTS -v`).

As an experiment, I also tried using continuations instead of the `Res` type (as is done in `parsec` and other parser combinator libraries), but that made compiling mhs a bit slower.

### DIFF
--- a/src/MicroHs/Parse.hs
+++ b/src/MicroHs/Parse.hs
@@ -3,6 +3,7 @@
 {-# OPTIONS_GHC -Wno-incomplete-uni-patterns -Wno-unused-do-bind #-}
 module MicroHs.Parse(P, pTop, pTopModule, parseDie, parse, pExprTop, keywords, dotDotIdent) where
 import qualified Prelude(); import MHSPrelude hiding ((*>))
+import Control.Applicative hiding ((*>))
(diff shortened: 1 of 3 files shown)


augustss:
This is a great fix.  I've been thinking about improving the parser, but never gotten around to it.

I also tried continuations, but it made it slower.  I wonder why.

PR #152, opened 2025-03-01
Make `Data.Char` report compliant

konsumlamm:
* Add `isLetter`, `isNumber`, `isMark`, `isSeparator`, `isLatin1`, `GeneralCategory`, `generalCategory`, `toTitle`, `showLitChar`, `lexLitChar`, `readLitChar`
  - Update `UniParse` to also generate a titlecase table
* Add `Storable Char` instance
* Make `chr` error on invalid codepoints
* Fix `toTitle`/`toUpper`/`toLower`: uppercase is different from titlecase, so e.g. `toUpper` must also convert titlecase letters


konsumlamm:
`toTitle`/`toUpper`/`toLower` are still not quite correct, since there are other codepoints that have title/upper/lower variants.

konsumlamm:
> `toTitle`/`toUpper`/`toLower` are still not quite correct, since there are other codepoints that have title/upper/lower variants.

To fix this, I changed them to simply always look up in the table. I also fixed `UniParse` to only generate consecutive ranges, which unfortunately makes the tables a lot bigger.

augustss:
I think that with a delta encoding of the to* tables there will be a lot of repeated table entries.  Which will at least compress well with LZ.  I had planned to do something like that, since what I originally had was onviously wrong, but more compact.

augustss:
I'll merge this, and we cam work on shrinking later.

augustss:
I've clawed back some of the performance losses for the new unicode tables.
The tables are now delta coded and the offset is zigzag encoded.

augustss:
The triples in the tables could be coded as a single Int.  It's easy for 64 bit ints, but looks a bit trickier for 32 bit ints.

PR #155, opened 2025-03-04
Fix error on GADTs with the result type in parens

meooow25:
I hope this is the right fix.

Without this change, the error is

```
"./GadtParen.hs": line 4, col 16: Bad GADT result type(((Foo a)),Foo,[(a$::_)])
```

because `dsGADT` gets `Nothing` from `getAppM`:

https://github.com/augustss/MicroHs/blob/7949c4bdd97ac05eff6e7d22fb99c349714d6df5/src/MicroHs/Parse.hs#L392-L398


augustss (commented):
Please move the test to where the other GADT tests are, instead of making a new one.

meooow25:
Anything I should do to make the emscripten tests pass?

augustss:
It makes no sense why that should fail. It's like it's using some old version of the compiler.

konsumlamm:
> Anything I should do to make the emscripten tests pass?

Regenerating `mhs.c` (using `make newmhsz`) should fix it.

meooow25:
Thanks, let me try that.

meooow25:
Fixed!

PR #162, opened 2025-03-08
Fix various UTF-8 functions

konsumlamm:
* Fix `Data.ByteString.isValidUtf8` to reject invalid codepoints and overlong encodings
* Fix `Data.Text.pack` to use the replacement character for surrogates
* Fix `System.IO.hPutChar` to error for surrogates
* Fix `System.IO.hGetChar` to error for surrogates
* Fix `System.IO.hGetContents` to error for surrogates
* Fix `getb_utf8` to error for overlong encodings


augustss:
I'm surprised that recompiling the compiler increased with >1% for this small change.

PR #166, opened 2025-03-09
Add the Bits Natural instance

### LINKED ISSUE #165: Add Bits Natural
Just like Bits Integer (#90). But there was a compile time concern there, so would this be alright?

meooow25:
Closes #165


meooow25:
I noticed that `isSigned :: a -> Bool` seems to be missing in Data.Bits compared to GHC's base. Is that on purpose? (Not that I have a need for it)

augustss:
The missing `isSigned` is not on purpose.  Feel free to add it

PR #168, opened 2025-03-10
Make emscripten CI less fragile

konsumlamm:
Run `make newmhs` before running the emscripten tests, to avoids failures like https://github.com/augustss/MicroHs/pull/155#issuecomment-2701373378.


augustss:
This assumes that GHC is available.  For many of the CI platforms there is no GHC.

PR #172, opened 2025-03-19
Add more functions to `Data.Text.Encoding`

konsumlamm:
Add encode/decode functions for UTF-16 and UTF-32 and a few other functions.


augustss (commented):
Really?  Latin1 has a one byte encoding for characters in the range 128-255, utf8 has two bytes, right?

konsumlamm:
> Really? Latin1 has a one byte encoding for characters in the range 128-255, utf8 has two bytes, right?

Yes, but those bytes are converted to `Char`s (which gives the correct codepoint), which are then `pack`ed into a `Text`.

PR #183, opened 2025-04-18
Allow operators in type constraints

konsumlamm:
Make `=>` a reserved operator (as in the Haskell report) and allow arbitrary operators in constraints.


augustss:
I'll fix the test.

PR #200, opened 2025-06-06
gmp fixes

claudeha:
some fixes to make mhs+gmp work on my debian bookworm linux system

probably mostly stale due to simultaneous changes, mainly posting in case something was missed


claudeha:
the simultaneously written changes in the master branch work for me, so this is probably uncessary

PR #218, opened 2025-06-10
Add TinyCC support

ikervagyok:
Added TinyCC for bootstrapping as well as directly running the latest `generated/mhs.c`.

Before merging the first commit (nix stuff) probably needs to be dropped.


augustss:
I've added some TinyCC support

PR #221, opened 2025-06-21
Add testforeign to the alltest target

TristanCacqueray:


augustss:
Can you update this so it works on your Linux system?

TristanCacqueray:
Sure, this now works on Fedora Linux x86-64.

PR #222, opened 2025-06-21
Add ffitest to the alltest target

TristanCacqueray:


TristanCacqueray:
Not sure what happened, but that part of #221 was not merged into master.

PR #248, opened 2025-07-30
provide an FFI wrapper for eval.c and make the  "mhs -r" feature work with GHC

thma:
This pull requests provides two enhancements that build up on another:

1. I created a small FFI Wrapper aroung eval.c in a new module `MicroHs.MhsEval`.
This module allows to run mhs combinator code (a la out.comb) from (ghc compiled) Haskell programs like so:

```haskell
import MicroHs.MhsEval (withMhsContext, run)

main :: IO ()
main = do

  combCode <- readFile "out.comb"
  withMhsContext $ \ctx -> do
    putStrLn "Running the compiled program:"
    run ctx combCode

```

2. In `MicroHs.Main` I used this feature to implement `mhs -r Example` for GHC compiled programs. 
With this you can use MHS as an embedded compile AND runtime environment. So in any GHC compiled program you can now do things like:

```haskell
import System.Environment (withArgs)
import MicroHs.Main (main)

main :: IO ()
main = do
  -- use MicroHs to compile AND execute the Example.hs program
  withArgs ["-r", "Example"] MicroHs.Main.main
```

I had to move `longUsage` out of MicroHs.main as the `\` -delimited multiline string interfered with the CPP preprocessor.


src/MicroHs/Main.hs
```diff
@@ -104,7 +105,7 @@ longUsage = usage ++ "\nOptions:\n" ++ details
       \-PPKG              Build package PKG\n\
       \-Q PKG [DIR]       Install package PKG\n\
       \-tTARGET           Select target\n\
-      \                   Distributed targets: default, emscripten, windows, tcc, environment\n\
```
augustss, line 107: Do a pull of the latest MicroHs sources.

src/MicroHs/MhsEval.hs
```diff
@@ -0,0 +1,141 @@
+{-# LANGUAGE ForeignFunctionInterface #-}
+{-# LANGUAGE CPP #-}
```
augustss, line 2: I don't want any CPP in MicroHs source files.
If different code is needed for different compilers then the module goes into `mhs/`, `ghc/`, and `hugs/`.
And the contents can the be compiler specific.  The build will use the correct `-i` flag to get the right module.

MicroHs.cabal
```diff
@@ -118,6 +119,11 @@ executable mhs
                        directory    >= 1.2 && < 1.6,
                        text         >= 2.0 && < 2.5,
                        array        >= 0.5 && < 0.6
+    include-dirs:      src/runtime src/runtime/unix
+    install-includes:  mhseval.h
+    c-sources:         src/runtime/mhseval.c
+    cc-options:        -DWANT_FLOAT64=1
```
augustss, line 125: Do you need this?  WANT_FLOAT64=1 is the default for the `unix` configuration.

src/MicroHs/Main.hs
thma: OK !

src/MicroHs/MhsEval.hs
thma: Ok, I'll change that. CPP is definitely a PITA

MicroHs.cabal
thma: OK, I'll crosscheck that.!

src/MicroHs/MhsEval.hs
thma: Done

thma:
Are there any things in my PR that you would like me to change or improve before pulling it?

augustss:
The code looks good.
I'm hesitating, because I'm not sure this feature is useful enough to be included.
The reason MicroHs is reasonably fast&small is that I have carefully weighed every new feature, and excluded some.

Let me test how much your code impacts compilation time.  If it's >0.1% I think it's too much.

thma:
Ok, that's a fair point.

augustss:
As it stands, recompiling the compiler is 1.5% slower.  But let me experiment a little.

augustss:
Compilation time increases with 0.095% with this mhs/MhsEval.hs.
That is acceptable. :)
```
module MhsEval (
    MhsContext,
    withMhsContext,
    run,
    compiledWithGhc
) where

compiledWithGhc :: Bool
compiledWithGhc = False

data MhsContext
type MhsCombCode = String

withMhsContext :: (MhsContext -> IO a) -> IO a
withMhsContext = err

run :: MhsContext -> MhsCombCode -> IO ()
run = err

err :: a
err = error "only supported with GHC"
```

augustss:
Can you update your code with the minimal MhsEval.hs?
And also pull the latest changes.

thma:
Great news! yes, I'll update the PR.

augustss:
There was an internal bfile API change.  All `_buf` functions have been renamed to `_mem`.

ghc/MhsEval.hs
```diff
@@ -0,0 +1,116 @@
...
+import Data.Typeable
+
+-- | Opaque context type
+newtype MhsContext = MhsContext (Ptr ())
+
+-- | Custom error type for MicroHs evaluation
+data MhsError = 
+    MhsInitError String
+  | MhsEvalError String
+  deriving (Show, Typeable)
```
augustss, line 27: Get rid of Typeable

PR #255, opened 2025-08-10
Add runtime foreign imports when compiling `emscripten`

HeinrichApfelmus:
This pull request adds

* A couple of foreign imports to the runtime when compiling with `emscripten`.
* A Makefile target `bin/mhseval.js` that is a compilation of `mhseval` to a JavaScript/WebAssembly runtime.

The foreign imports are implemented in `mhseval.js`. They are powerful enough to call any JavaScript function from Haskell space (at a performance cost due to `eval`), and to use Haskell functions as event handlers in JavaScript space.

To use MicroHs in the web browser, add a combinator file `out.comb` with the Emscripten file system and add a `<script src="mhseval.js"></script>` script tag to execute the combinator file.

MicroHs in the web browser, here we come! `⟩λ=`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,14 @@ bin/mhsevalsane:	$(RTS)/*.c $(RTS)/*/*.h
 	@mkdir -p bin
 	$(CCEVAL) $(CCSANITIZE) $(RTSINC) $(RTS)/comb.c $(CCLIBS) -o bin/mhsevalsane
 
+# mhseval, compiled with emscripten. Use in the browser!
+EMCCOPTS= -O3 -sEXPORTED_FUNCTIONS=_apply_sp,_main -sEXPORTED_RUNTIME_METHODS=stringToNewUTF8 -sALLOW_MEMORY_GROWTH -sTOTAL_STACK=5MB -sSINGLE_FILE -DUSE_SYSTEM_RAW -Wno-address-of-packed-member
+EMCCLIBS= -lm
+EMCCEVAL= emcc $(EMCCOPTS) $(RTSINC) $(MAINC) $(RTS)/eval.c
+bin/mhseval.js:	$(RTS)/*.c $(RTS)/*.h $(RTS)/*/*.h
+	@mkdir -p bin
+	$(EMCCEVAL) $(RTS)/comb.c $(EMCCLIBS) -o bin/mhseval.js
+
 # Compile mhs with ghc
 bin/gmhs:	src/*/*.hs ghc/*.hs ghc/*/*.hs ghc/*/*/*.hs
 	@mkdir -p bin


HeinrichApfelmus:
We should probably add documentation about how to use `mhseval.js` somewhere.

augustss:
If you document how to use mhseval.js, I'll include it in the README (which I plan to reorganize).

Can one run threepenny-gui programs with microhs now?

HeinrichApfelmus:
> If you document how to use mhseval.js, I'll include it in the README (which I plan to reorganize).

👍 

> Can one run threepenny-gui programs with microhs now?

Not yet — I need to polish my changes to the FFI logic and go through the `threepenny-gui` modules in order to remove any stuff that cannot be compiled with MicroHs. But that should be busywork on my part at this point.

That said, there is one part where I need help from the runtime: garbage collection, in particular `System.Mem.Weak`. In particular, I need the precise semantics where the construction of a weak pointer introduces a link between key and value, and we can forget about the `Weak` pointer object entirely. It's ok for the key to a thing with identity; specifically, I'm using a function
```
mkWeakIORefValue :: IORef a -> value -> IO () -> IO (Weak value)
```

The code will still run without `System.Mem.Weak`, of course — but it will leak memory… I have some fancy setup for cross-runtime garbage collection, where Haskell takes ownership of JavaScript objects.

augustss:
OK.  I'll look at implementing weak pointers.  I've not given it any thought yet.

augustss:
I guess that you want `mhseval.js` to be part of the distribution so people are not forced to install emscripten?
This is not a problem, I'll just put it in `generated/` and build it as part of making a new version.

HeinrichApfelmus:
> I guess that you want mhseval.js to be part of the distribution so people are not forced to install emscripten?
> This is not a problem, I'll just put it in generated/ and build it as part of making a new version.

Yes, that would be swell! 😊

augustss:
I've looked at ghc's weak pointers. It doesn't look very difficult to implement. But it's another complication in the runtime. 
I'll see if I can get it done later this week.

HeinrichApfelmus:
> I need to […] go through the `threepenny-gui` modules in order to remove any stuff that cannot be compiled with MicroHs. But that should be busywork on my part at this point.

It compiles and runs! 😱 Some of the examples work, e.g. I get an interactive HTML page for  `CurrencyConverter.hs`. Still needs some work under the hood and some convenience around building the HTML page, but it's working.

augustss:
Wow!  Very cool!

augustss:
Can I try it myself?

HeinrichApfelmus:
Uh, sure. It's not open for business just yet, but you will get a result by following these steps:

Install dependencies

* `mcabal install` from the `master` branch at `https://github.com/HeinrichApfelmus/vault`
* `mcabal install transformers`
* Check out the branch `HeinrichApfelmus/preview` from https://github.com/HeinrichApfelmus/threepenny-gui
  * `mcabal install` the package in `lib/javascript-ffi-threepenny`
  * `mcabal install` the package in `lib/threepenny-gui`

Compile example

* `cd` to `lib/threepenny-gui/samples/MicroHs` and execute `build-microhs.sh`.
  * This should produce a file `Example.js`
  * Run it by opening `ExampleMicroHs.html`, assuming that the browser allows local files

If you replace the contents of `lib/threepenny-gui/samples/MicroHs/Example.hs` with the contents of `lib/threepenny-gui/samples/CurrencyConvert.hs`, then you should get a different example.

augustss:
It works!!  That's so cool.  Thank you for doing this.
The JS files are not very small, but it's not terrible.

augustss:
I've implemented a crappy `mdo`.  (It slows down the compiler by 0.2%.)

HeinrichApfelmus:
> I've implemented a crappy mdo. (It slows down the compiler by 0.2%.)

Thanks! 😊 This does the thing nicely, my `CRUD.hs` example now compiles and works out of the box.

PR #261, opened 2025-08-24
Make `Numeric` report compliant

konsumlamm:
Make `Numeric` report compliant by exporting more functions that were already there and newly defining `readFloat` and `fromRat`. The `readFloat` code is taken verbatim from GHC base, the `fromRat` implementation is adapted from the GHC version. `Numeric` now imports `Prelude`, since I didn't want to write down all the needed imports, I hope that's okay .

### DIFF
--- a/lib/Numeric.hs
+++ b/lib/Numeric.hs
@@ -57,52 +57,38 @@ fromRat (n :% d) | n > 0     = fromRat' (n :% d)
                  | n < 0     = - fromRat' ((-n) :% d)
                  | otherwise = encodeFloat 0 0             -- Zero
 
--- Conversion process:
--- Scale the rational number by the RealFloat base until
--- it lies in the range of the mantissa (as used by decodeFloat/encodeFloat).
--- Then round the rational to an Integer and encode it with the exponent
--- that we got from the scaling.
--- To speed up the scaling process we compute the log2 of the number to get
--- a first guess of the exponent.
-
 fromRat' :: (RealFloat a) => Rational -> a
 -- Invariant: argument is strictly positive
 fromRat' x = r
-  where b = floatRadix r
-        p = floatDigits r
-        (minExp0, _) = floatRange r
-        minExp = minExp0 - p            -- the real minimum exponent
-        xMax   = toRational (b ^ p)
-        ln     = integerLogBase b (numerator x)
-        ld     = integerLogBase b (denominator x)
-        p0     = (ln - ld - p) `max` minExp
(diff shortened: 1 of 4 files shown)


lib/Numeric.hs
```diff
@@ -1,14 +1,113 @@
...
+        -- if ln - ld >= minExp0, then b^(p-1) < x0 < b^(p+1), so there's at most
+        -- one scaling step needed, otherwise, x0 < b^p and no scaling is needed
+        (x', p') = if x0 >= xMax then (x0 / toRational b, p0+1) else (x0, p0)
+        r = encodeFloat (round x') p'
+
+wordLog2 :: Word -> Int
+wordLog2 w = _wordSize - 1 - primWordClz w
+
+integerLog2 :: Integer -> Int
+integerLog2 (I sign ds) =
```
augustss, line 92: This won't work if you compile to use GMP.  The internal representation of Integer should never be used.
Perhaps the integerLog stuff should move to some other module.

I also think that the fromRat code is much more complex than necessary (I don't mind saying that, since I'm the original author of the code, almost 35 years ago).  I think a variation of `fromRational` from `Fractional Double` would be better.

konsumlamm: > This won't work if you compile to use GMP.

Right, of course. There should probably be a CI job using GMP.

> I think a variation of `fromRational` from `Fractional Double` would be better.

Then I have some questions on [that code](https://github.com/augustss/MicroHs/blob/e72d01f5be926255fe25320f2fec83bdc90ce919/lib/Data/Double.hs#L49-L64):
- Why 1023 instead of 1024 (which is the maximum exponent)?
- Why require `d < maxVal` instead of `d <= maxVal`?
- How was 256 chosen as shift?

I don't think anyone really uses `fromRat` (most people probably don't even know that it exists), so it's not important for it to be very efficient. I'm happy to go with your suggestion, if that's what you prefer.

augustss: Yes, a CI job for GMP would be great.  It really only has to be run on one platform.

Answers: all the answers are that it was just accidental.  It doesn't really matter exactly what the constants are, as long as they scale numbers that are too big.  It should be possible to figure out some better set of constants from the `RealFloat` information.  In which case `Double` and `Float` could share a `fromRational` implementation.

konsumlamm:
I removed the `integerLog2` special case and added a shortcut, if the numerator and denominator fit in a `RealFloat a => a`. I'll see if an implementation based on `fromRational @Double` works better.

PR #263, opened 2025-08-26
Fix GMP conversion functions

konsumlamm:
`mpz_get_ui` ignores the sign, so it gets the least significant bits of the absolute value, which is not the behaviour we want. Instead, I modified `mpz_get_si` to properly handle `maxBound :: Int`.
Similarly, I changed `mpz_get_si64` to use the code of `mpz_get_ui64`, with the small change to use `mpz_tdiv_q_2exp` (logical right shift) instead of `mpz_fdiv_q_2exp` (arithmetic right shift), to get the absolute value (which is then negated if the input was negative).
`mpz_init_set_si64` was also broken for negative values, so I changed it to use `mpz_init_set_ui64` and negate the result if the input was negative.

### DIFF
--- a/lib/gmp/Data/Integer_Type.hs
+++ b/lib/gmp/Data/Integer_Type.hs
@@ -29,8 +29,7 @@ newtype Integer = I (ForeignPtr MPZ)
 foreign import capi "new_mpz"         new_mpz         :: IO (Ptr MPZ)  -- it really returns a ForeignPtr
 foreign import capi "mpz_init_set_si" mpz_init_set_si :: Ptr MPZ -> Int -> IO ()
 foreign import capi "mpz_init_set_ui" mpz_init_set_ui :: Ptr MPZ -> Word -> IO ()
---foreign import capi "mpz_get_si"      mpz_get_si      :: Ptr MPZ -> IO Int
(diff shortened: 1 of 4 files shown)


augustss:
Can you add some tests that cover these cases?

konsumlamm:
Sure, will do.

konsumlamm:
I added tests for `fromInteger`/`toInteger` (which use the functions i modified under the hood) and they pass, but now the bootstrap test is broken again. `Integer`s on 32 bit systems (without GMP) also seem to be broken.

augustss:
Just adding the tests makes it fail on 32 bit.  I'm debugging that.

augustss:
I've fixed the 32 bit bug.  And I've merged the tests.

augustss:
I'll look at the bootstrap failure later.
If you want to try, here's what I recommend:
compile with `bin/mhs -ddump-linked` and `bin/mhs-stage1 -ddump-linked` and `diff` the outputs.  This will tell you where the difference is.

konsumlamm:
I found the error, `mpz_get_si64` on 64 bit platforms was still using `mpz_get_si`.

augustss:
Thanks for finding and fixing the bugs!

PR #265, opened 2025-08-27
Implement deriving for `Ix`

konsumlamm:
Implement deriving for `Ix`, as specified in [the report](https://www.haskell.org/onlinereport/haskell2010/haskellch19.html#x27-22700019.2). While the implementation seems to work, there are some things I'm unsure about:
- Is it fine to reuse `EVar`s in multiple definitions? I assume so, but I want to make sure.
- For the case of only having nullary constructors, I create helper functions for converting between the type and `Int`. To make the names unique, I currently append `"@" ++ show lhs`. Is there a better way to get a unique name here? And is it fine to return the function definitions with the instance?

And if you have other suggestions for improvement, please let me know!

### DIFF
--- a/src/MicroHs/Deriving.hs
+++ b/src/MicroHs/Deriving.hs
@@ -280,7 +280,7 @@ derEnum mctx 0 lhs cs@(c0:_) enm | all isNullary cs = do
           x1 = EVar (mkIdentSLoc loc "x1")
           x2 = EVar (mkIdentSLoc loc "x2")
         in eEqn [x1, x2] (EIf (eAppI2 (mkBuiltin loc ">=") (EApp (EVar iFromEnum) x2) (EApp (EVar iFromEnum) x1)) (eAppI3 iEnumFromThenTo x1 x2 eLastCon) (eAppI3 iEnumFromThenTo x1 x2 eFirstCon))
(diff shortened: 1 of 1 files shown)


augustss:
- It's perfectly fine to reuse `EVar`s
- There's no nice way to generate unique names, since there is no monad used for deriving.  If your code works, then I guess it was fine to return functions together with the generated instance.  It's not something I have thought about.  I'm not thrilled about generating the extra functions.  I'll think a little bit about if there is another way.  Can you reuse the function for deriving `Enum` for your `toInt` and `fromInt`?

konsumlamm:
> Can you reuse the function for deriving `Enum` for your `toInt` and `fromInt`?

Indeed, the functions are basically the same, in fact, I just copied the implementation of `fromEnum`/`toEnum`. I factored them out.

EDIT: CI fails due to an SSL timeout.

augustss:
I've made a small change so that derived instances can have extra bindings in the body.
So now you can place `fromInt` and `toInt` among the method bindings instead of being global.

I have often wished extra bindings was not an error in Haskell, but we would need some syntax to indicate that the extra bindings are intentional.  Maybe an extra `where`, e.g.
```
instance C T where
  m x = foo x + 1 -- method binding
 where
  foo x = ...     -- extra binding
```

augustss:
You need a small update: `Instance ...` should be replaced by `Instance ... extra` where `extra` are the non-method bindings.

augustss:
Thanks!

PR #266, opened 2025-08-28
Implement deriving for `Read`

konsumlamm:
Implement deriving for `Read`. The generated instances are based on what GHC generates, rather than what's described in the report. Also fix the derived `Show` instance for records.

The derived instances aren't completely report compliant yet, since constructors that were written in infix form should also be read/shown in infix form, respecting the precedence. However, this information is not available to the derivers.


augustss:
To avoid circular imports and to avoid code bloat with `Read` instances, it might be a good idea to remove the `deriving Read` from the types in base and instead use standalone deriving for those in `Text.Read`

augustss:
It's a shame about fixity.  The fixity table is not built until after the deriving code has run.  I'll give that some thought.

konsumlamm:
> To avoid circular imports and to avoid code bloat with `Read` instances, it might be a good idea to remove the `deriving Read` from the types in base and instead use standalone deriving for those in `Text.Read`

Which instances do you mean? `Data.Functor.{Compose, Product, Sum}` already import `Prelude` and `Void` and `Natural` are barely used. Moving the `Array` instance should work.

augustss:
It's awesome that you have implemented `deriving Read`.  It has been on my TODO list forever.

I'm just about to commit a change so that `Typeable` is always derived, just like ghc.  It sadly adds 1.5% to the time to recompile the compiler.

augustss:
BTW, don't try deriving `Data`.  There's a compiler problem I need to fix first.

augustss:
I was wrong.  The fixity is available when deriving!
Look at `_fixLookup` in `DerShow`.  So far unused.

PR #267, opened 2025-08-29
Make `Data.Complex` (more or less) report compliant

konsumlamm:
- Add a `Read` instance, based on what GHC would generate (since deriving wouldn't produce the correct instance yet)
- Fix the `Show` instance to consider precedence
- Add an export list
- Uncomment the `**` definition
- Use literal patterns (I assume the reason they were commented out is that they weren't supported when the module was added?)

With these changes, `Data.Complex` is (more or less report) compliant. Technically, the report has a `RealFloat a =>` datatype context, so every instance also requires that (and the `Show` instance doesn't require `Show a`). However, datatype contexts are widely considered a misfeature; they're deprecated in GHC and MicroHs doesn't support them at all. GHC just derives `Read` and `Show` (which generates the instances I added), so we should be good.

### DIFF
--- a/lib/Data/Complex.hs
+++ b/lib/Data/Complex.hs
@@ -1,16 +1,36 @@
-module Data.Complex(module Data.Complex) where
+module Data.Complex (
+  Complex((:+)),
+  realPart,
+  imagPart,
+  mkPolar,
+  cis,
+  polar,
+  magnitude,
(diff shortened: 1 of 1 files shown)


augustss:
Thanks!  This module was indeed quite old.

Issue #278, opened 2025-09-18
Importing data types using `{-# SOURCE #-}` imports doesn't import constructors


konsumlamm:
### Minimal example
`Module.hs`:
```hs
module Module (Data(..)) where
data Data = A | B | C deriving Show
```
`Module.hs-boot`:
```hs
module Module where
data Data
```
`Main.hs`:
```hs
import {-# MODULE #-} Module
main = print A
```
This fails to compile with the error `undefined value: A`.
### Effects
For example, the constructors of `GeneralCategory` aren't exported by `Data.Char`, because it uses a `{-# SOURCE #-}` import for `Data.Char.Unicode`.

augustss:
Thanks for the report. I did actually notice this myself just a couple of days ago.

augustss:
I think this is fixed now.  33a1b7c139e7b6cc282fba0656c2f3a0581a095f

konsumlamm:
The example above still fails to compile with the same error.

augustss:
Should it compile? The -boot file does not list the constructors, so I don't think it should. 
But there was a bug before so even if you had the constructors in the boot file it would not work.

konsumlamm:
I think so, at least GHC accepts it. The [GHC User's Guide](https://downloads.haskell.org/ghc/latest/docs/users_guide/separate_compilation.html#mutual-recursion) says
> A hs-boot file need only contain the bare minimum of information needed to get the bootstrapping process started. For example, it doesn’t need to contain declarations for everything that module A exports, only the things required by the module(s) that import A recursively.

augustss:
But the -boot module does not contain everything Main needs. Main uses A, and A is not exported by the -boot module. The point of the boot module is that it contains all the definitions needed in the module that does the SOURCE import.

konsumlamm:
But it does contain everything needed to break the recursive cycle. In fact, since there is no cycle here, it even works if `Module.hs-boot` only contains
```hs
module Module where
```

augustss:
If you make it recursive (`Module` importing `Main`) it doesn't compile in GHC either.
The `-boot` module does not contain enough information to break the cycle. `Main` needs `A`, how can it get hold of `A`?  Only by compiling `Module.hs`, but `Module.hs` depends on `Main.hs`, so that's not possible.
It can only compile `Module.hs-boot`, which contains no information about A.

The way `-boot` modules work in MHS (and GHC) this should not work.

[...]

PR #291, opened 2025-09-24
Add missing `Ix` instances

konsumlamm:
The instance for `SeekMode` is commented out, since `SeekMode` doesn't exist yet.

The instance for `GeneralCategory` currently doesn't compile, because of #278. I can wait for a fix, try to find a workaround or comment it out for now. Which option do you prefer?


augustss:
I'm going to fix #278 in the next few days.

PR #292, opened 2025-09-25
implement ShortByteString operations matching API of bytestring-0.12.2.0

claudeha:
implement `ShortByteString` operations in terms of `ByteString` operations

`ShortByteString` is used by `os-string`.

Currently missing `createFromPtr` and `copyToPtr`.


claudeha:
good point about coerce, it completely slipped my mind.  I'll modify this PR soon...

PR #296, opened 2025-09-28
Build more packages in CI

konsumlamm:
`split` and `monad-loops` both have no dependencies apart from `base`, are pretty popular, and build with MicroHs out of the box. Let's make sure that it stays that way!

### DIFF
--- a/.github/workflows/hackage-ci.yml
+++ b/.github/workflows/hackage-ci.yml
@@ -128,6 +128,30 @@ jobs:
         cd granite
         mcabal install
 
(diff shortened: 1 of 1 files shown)


augustss:
I'm sure there are more packages that build.

konsumlamm:
Do you have a policy which packages should be included in the CI?

augustss:
No policy.  I've just added packages as I've been fixing them and using them.
Any package that is even moderately useful and which compiles is welcome.

PR #297, opened 2025-09-29
Derive `Read` instances for `Data.Bits` newtypes

konsumlamm:
Is there any reason the `Read` derives were commented out? Deriving `Read` works since #266 and the required imports are available from `Mhs.Builtin`, which is already imported.


augustss:
Thanks!  I had them commented out during reorganization of Data.Bits

Issue #290, opened 2025-09-24
import qualified Prelude makes (:) go away


claudeha:
ghc still has it:

```
~/code $ cat bug.hs                                               import qualified Prelude as P
main = case "test" of                                                 
  (a:b:c) -> P.return ()                                          
~/code $ ghc bug.hs                                               Loaded package environment from /data/data/com.termux/files/home/.ghc/aarch64-linux-9.12.2/environments/default                     
[1 of 2] Compiling Main             ( bug.hs, bug.o )             
[2 of 2] Linking bug                                              
~/code $ mhs bug.hs
mhs: uncaught exception: error: "bug.hs": line 4, col 5: undefined value: :          
```

cannot import it in ghc (not exported), nor mhs (syntax error)

```                                               
~/code $ cat bug2.hs                                              
import qualified Prelude as P
main = case "test" of                                             
  (a P.:b P.: c) -> P.return ()                                   
~/code $ ghc bug2.hs                                              
Loaded package environment from /data/data/com.termux/files/home/.ghc/aarch64-linux-9.12.2/environments/default                     [1 of 2] Compiling Main             ( bug2.hs, bug2.o )           bug2.hs:4:6: error: [GHC-76037]                                       
Not in scope: data constructor ‘P.:’                              
Note: The module ‘Prelude’ does not export ‘:’.                 |                                                              
 4 |   (a P.:b P.: c) -> P.return ()      
                           |      ^^^
bug2.hs:4:11: error: [GHC-76037]                                      
Not in scope: data constructor ‘P.:’                              
Note: The module ‘Prelude’ does not export ‘:’.                 |                                                               
4 |   (a P.:b P.: c) -> P.return ()                           
      |           ^^^
~/code $ mhs bug2.hs
mhs: uncaught exception: error: "bug2.hs": line 4, col 16:          found:    )                                                       
expected: { . LQIdent ( UQIdent [ literal _primitive @ \ ...
```

this came up trying (failing so far) to get filepath package working

augustss:
As much as I dislike making (:) special, I'm doing it.
40150e78981247129e67fe9a7789ca243e756d1e

claudeha:
similarly for empty list literal (non-empty lists work fine)

```
~/code $ cat list.hs
import Prelude (return)

main = case [] of [] -> return ()
~/code $ ghc list.hs
[1 of 2] Compiling Main             ( list.hs, list.o )
[2 of 2] Linking list
[...]

PR #304, opened 2025-10-03
Fix desugaring of `[]` in types so that it can be used without imports

konsumlamm:
I hope this is the correct fix, I did some experimenting and this seems to work.

I modified the `ImpMet` test, since it already only imports specific symbols from `Prelude`, but I can also move this to a new test if you prefer. It also tests the use of `:` and `[]` in expressions (#290).


augustss:
The right fix is to add the line
       `(mkIdentB "[]",           [entry identList     kTypeTypeS]),`
in `primTypes`

PR #307, opened 2025-10-07
Update bytestring modules

konsumlamm:
The `Char8` modules defined their own `ByteString` type, but in `bytestring`, they use the existing one. I added most functions from `bytestring`, mostly using the original implementations. I also added some functions that were needed to `Data.ByteString.Lazy`. The `intToWord8` & `word8ToInt` helpers were moved to `Data.Word.Word8`, so that they can be used for defining `c2w` & `w2c` in `Data.ByteString.Internal`.


augustss:
Thanks!

Interesting, I also made Char8 no longer be its own type, but I've not pushed that.

PR #312, opened 2025-10-20
dealing with packages with external FFI dependencies

arossato:
With regards to issue #311 here's the PR.

thank you!

Issue #311, opened 2025-10-20
Library packages linking to external libs via FFI


arossato:
Presently mhs installs packages as serialized Package records, while mcabal, when a "c-sources" field is used, will install the corresponding C files in the `data` directory of the package and `mainCompileC` will use them to produce a correctly linked executable.

If a package links to an external library via the FFI, then everyone using the library will need to pass to mhs a flag, through the `-optc` option, to specifically link the exec with the external library.

To illustrate the issue I created a simple demo:

https://codeberg.org/andrea_rossato/microhs-ffi-libs

Using lib1 and lib2 is not a problem, since no external libraries are linked and everything is available to mhs. So, after installing the libraries with `mcabal install`, when compiling `lib1_main.hs` and `lib2_main.hs` mhs will produce the correctly linked binaries. But with lib3 and lib4, which are linking to the glibc `libcrypt` library, when compiling the corresponding `lib3_main.hs` and `lib3_main.hs` will generate a linker error.

To solve the issue I came up with the idea of:

- adding an `-optl` flag for mhs, for passing to the C compile flags for the linker;

- adding to the `Package` record a `pkgTarget` and a `pkgOptl` field, for storing the target and the linker options used for compiling the library (the same option is present in ghc);

- moreover, in order to check if the package compiles and correctly links to the external library, when `-optc` or `-optl` options are used mhs will try to compile a stub executable.

This way when creating a package mhs will know how to link the executable which is using the library.

With these changes all the examples of my demo compile correctly.

I created a branch in my microhs fork to show my approach:

https://github.com/arossato/MicroHs/tree/linker

I've also added support for the new `-optl` flag in mcabal. This is done with the `extra-libraries` field.  I pushed to my mcabal fork the code to support this change:

https://github.com/arossato/MicroCabal

I'd like also to add support for the  pkgconfig-depends field.

If you are willing to accept these changes I can create pull requests for them. Please let me know.

PS: I've also added a `-keep-tmp-files` flag to mhs (the same is present in ghc) to keep temporary files. I think this may be useful for debugging purposes.

augustss:
Thanks! I will take a look and merge it into mhs/mcabal.

augustss:
Please create a PR.

arossato:
done. Thank you!

arossato:
I think this should be closed as completed.

PR #313, opened 2025-10-25
Add a linker flag for packages with external FFI dependencies

arossato:
amended version of #312 for dealing with issue  #311

### DIFF
--- a/src/MicroHs/Main.hs
+++ b/src/MicroHs/Main.hs
@@ -234,6 +234,20 @@ mainBuildPkg flags namever amns = do
   t2 <- getTimeMilli
   when (verbose flags > 0) $
     putStrLn $ "Compression time " ++ show (t2 - t1) ++ " ms"
+  unless (null (cArgs flags) && null (lArgs flags)) $ do
+    -- we are linking to some c lib so we test if the package compiles
+    when (verbose flags > 0) $
+      putStrLn $ "Testing the library " ++ namever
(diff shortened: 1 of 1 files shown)


src/MicroHs/Package.hs
```diff
@@ -30,6 +30,8 @@ data Package = Package {
   pkgName      :: IdentPackage,                    -- package name
   pkgVersion   :: Version,                         -- package version
   pkgCompiler  :: String,                          -- compiler version that created the package
+  pkgTarget    :: String,                          -- copilation target
```
augustss: I don't want the target in the package. Packages are mostly target independent.

src/MicroHs/Main.hs
```diff
@@ -229,6 +234,20 @@ mainBuildPkg flags namever amns = do
   t2 <- getTimeMilli
   when (verbose flags > 0) $
     putStrLn $ "Compression time " ++ show (t2 - t1) ++ " ms"
+  unless (null (cArgs flags) && null (lArgs flags)) $ do
```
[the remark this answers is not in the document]
arossato: I understand and I agree. Depending on your answer to my previous replay I will amend this PR according to your suggestions.

src/MicroHs/Package.hs
arossato: maybe Flags.lArgs :: (String, [String])

so that only lArgs are target specific?

Issue #315, opened 2025-10-31
Lens support (micolens specifically)


TCCQ:
I'm really interested in trying to get mhs to be more than a toy compiler, since I think the simplicity and compactness of the ideas is refreshing compared to the rest of the haskell ecosystem.

To the end, I would love to be able to compile any of the haskell I use or could imagine using using mhs. From some poking around one blocker on building many existing projects is the lack of lenses. 

Microlens seems to be a feasible target as it is relatively compact, has some real uses and doesn't require template haskell. At the moment it doesn't seem to compile (I think because of a lack of TypeOperators or possibly a lack of fully qualified kinds ?). 

Does anyone have any ideas about what would need be done to get this working, on the microlens or the microhs side? I'm interested in trying to get it off the ground, but I'm not really sure what I am looking for.

augustss:
MicroHs certainly supports Type operators. I'm not sure what you mean by fully qualified kinds. Do you mean a way of making new kinds? While not supported, it's something I want to add. 

I'll take a look at microlens.

konsumlamm:
The reason `microlens` doesn't compile is because it uses type families.

augustss:
I just looked at microlens. It looks like the type family stuff could be ifdef:ed out and it would still retain most of the functionality. 

I need to implement type families soon. The lack is holding MicroHs back.

arossato:
I don't know if this may be useful but lenses could be implemented in mhs with rank2Types.

This works:

```` haskell
module Lens where

import Control.Applicative  ( Const (..) )

-- | Identity functor
newtype Identity a = Identity { runIdentity :: a }

-- | Functor instance
instance Functor Identity where
  fmap f (Identity a) = Identity (f a)

-- | Type alias for lens
type Lens s a = forall f. Functor f => (a -> f a) -> s -> f s

-- | Viewing
view :: ((a -> Const a a) -> s -> Const a s) -> s -> a
view l s = getConst (l Const s)

-- | Setting
set :: ((a -> Identity b) -> s -> Identity t) -> b -> s -> t
set l b = runIdentity . l (\_ -> Identity b)


-- lens :: forall a s . (s -> a) -> (s -> a -> s) -> Lens s a -- not accepted by mhs
lens :: forall a s f . Functor f => (s -> a) -> (s -> a -> s) -> (a -> f a) -> s -> f s
lens sa sas aa s = sas s <$> aa (sa s)

data A = A {f :: Int} deriving Show

fL :: Lens A Int
[...]

PR #316, opened 2025-11-03
process the constraints and do not skolemize type variables in nested forall

### LINKED ISSUE #208: Nested foralls: Cannot satisfy constraint: ((=>) ~ (->))
Test case:

```haskell
incu :: () -> (forall a. Num a => a -> a)
incu _ x = x + 1

main = print (incu () 41)
```

Expected behavior (ghc used as reference)

```
$ runhaskell Test.hs
42
```

Actual behavior (mhs)

```
$ ./bin/mhs -r Test.hs

mhs: uncaught exception: error: "Test.hs": line 2, col 8: Cannot satisfy constraint: ((=>) ~ (->))
     fully qualified: (Primitives.~ Primitives.=> Primitives.->)
```

arossato:
addressing issue #208

Since there seems to be some pressure to fix #208 (see also #315) I started digging into it and I think I now have an idea of what is going on.

```` haskell
incu :: forall a . Num a => a -> (forall b. (Num b, Integral b) => b -> a)
incu x = (+) x . fromIntegral

main = print (incu 1 41)
````

This code would produce unsolved equality constraints. As far as I can understand when `incu` pattern matching is type-checked against its type signature, the forall part will eventually be processed by `unArrow` (via `tcExprR -> tcEqn -> tcPats`) in order to check each pattern type.

In `unArrow` the type part of the `Eforall _ _ t` will go through:

```` haskell
-- XXX Is this correct?
  (_, t') <- shallowSkolemise iks t
  unArrow loc t'
````

I think it is not correct: shallowSkolemise will generate skolems, which cannot be unified. That will create constraints that cannot be satisfied. Instead, if I understand it correctly, the correct thing to do is:

```` haskell
  t' <- tInstForall iks t
````

Now, I think this should be done *before* calling unArrow, as you will see, because everytime unArrow will be called for inner foralls, those parts will be re-instantiated and unification with the outer forall will be impossible.

So, my sort of fix, which works with the above example, doesn't work for those:

```` haskell
incu' :: forall a . (Num a, Show a) => a -> forall b. (Num b, Integral b) => b -> forall c . Show c => (a -> c) -> String
incu' x y f =
  let n = x + fromIntegral y
  in show $ f n

incu'' :: forall a . (Num a, Show a, Integral a) => a -> forall b. (Num b, Integral b) => b -> forall c . Show c => (b -> c) -> String
incu'' x y f =
  let n = x + fromIntegral y
  in show $ f (fromIntegral n)
````

(sorry for this stupid code but I was in a hurry to find something to test which works with ghc)

Even if I do not have a fix (I still have to figure out when tInstForall should be called), I thought it could be useful to share what I've done so far. Maybe some suggestions may come.

ps: the commit adds also the code for dealing with `=>`

### DIFF
--- a/src/MicroHs/TypeCheck.hs
+++ b/src/MicroHs/TypeCheck.hs
@@ -2289,7 +2289,12 @@ unArrow loc (EForall _ iks t) = do
   -- Found forall in a co-variant position.
   -- Make new unique tyvars in case of clashes.
   -- XXX Is this correct?
-  (_, t') <- shallowSkolemise iks t
+  t' <- tInstForall iks t
+  unArrow loc t'
+-- handle =>
+unArrow loc t | Just (ctx, t') <- getImplies t = do
+  d <- newDictIdent loc
+  addConstraint d ctx
   unArrow loc t'
 unArrow loc t = do
   case getArrow t of


augustss:
Thanks for trying to fix the problem, but I don't think your fix does the right thing. 

I know what's wrong, but fixing it requires some restructuring of the existing code.

arossato:
I suspected you knew the problem and the solution, but I thought it would be an enjoyable learning experience to try to understand what was going on and dig a bit into the 'beast', and nice having your feedback.

Thanks for sharing this project.

PR #320, opened 2025-11-13
Add instance of `Enum` for `IOMode`

MaximilianAlgehed:
This is in `ghc` but not in `mhs` and it caused me to have to write extra code in QuickCheck (boo!)


augustss:
Thanks for fixing it. It was an oversight.

PR #333, opened 2025-12-07
Fix runtime sources missing from Cabal package

AlexandreTunstall:
When trying to package MicroHs in Nixpkgs, I noticed that some sources are missing from the output directory.
As a result, the necessary files were not installed in the `CABALDIR` and the build output was broken.

### DIFF
--- a/MicroHs.cabal
+++ b/MicroHs.cabal
@@ -33,14 +33,19 @@ extra-source-files:
       paths/Paths_MicroHs.hs
       src/runtime/*.c
       src/runtime/*.h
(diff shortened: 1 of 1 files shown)


augustss:
Building MicroHs with cabal is a bad idea. The compiler you get is generated by ghc so it will miss some features. MicroHs should always be built with `make install`.

AlexandreTunstall:
> The compiler you get is generated by ghc so it will miss some features.

I am building it with MicroHs and MicroCabal, so it should be fine. Aside from this, it's been working fine so far.

PR #336, opened 2025-12-07
Add a reusable GitHub Action that installs MicroHs

sol:
This PR provides a GitHub Action to install a "nightly" build of MicroHs for CI use by third-party packages.

- This reduce e.g. the [QuickCheck MicroHs CI](https://github.com/nick8325/quickcheck/pull/454)  time by -2 minutes.
- It also lowers the barrier to add MicroHs CI jobs to more projects.
- Right now this still uses binaries from my fork, this is why it is a draft PR for now.  I have a separate PR in the pipeline to remedy this.

I also added instructions to the README on how to set up CI for a third-party package.  I don't care about the exact wording so please feel free to change things as you see fit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -519,9 +519,42 @@ Contributions are very welcome!
 When modifying the compiler, run `make newmhs` or `make newmhsz` (the latter compresses the binary)
 to generate a compiler that includes your changes.
 
-## Libraries
+## Third-party packages
 
-The libraries live in the `lib/` directory. Adding missing functions/instances/types from the report is a welcome contribution.
+Adding MicroHs support to a third-party package may take some initial effort, either by modifying the package itself, or by extending MicroHs.
+
+- For conditional compilation you can test for `__MHS__` via CPP.
+
+Once a package compiles with MicroHs, make sure it stays that way by adding a GitHub Action to the project repository:
+
+```yaml
+# file .github/workflows/mhs.yml
+
+name: MicroHs
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install MicroHs
+        uses: augustss/MicroHs@nightly
+
+      - name: Clone project repository
+        uses: actions/checkout@v4
+
+      - name: Build project
+        run: mcabal -r build
+```
+
+## Bundled libraries
+
+The bundled libraries live in the `lib/` directory. Adding missing functions/instances/types from the report is a welcome contribution.
 Common things from `base` and GHC boot libraries that use a lot of GHC-specific code (such as `array`, `bytestring`, `text`, ...) can also be added.
 
 ## Tests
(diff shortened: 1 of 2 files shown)


augustss:
I like the idea, but I have some issues with the execution of it.
I think the bundle that should be produces nightly should be the files in `generated/`, `src/runtime`, together with `base.pkg`.  Installing them is then a matter of compiling the three `.c` files, copying the runtime, and having `mhs` install the `base` package.

I've been planning to make a `Makefile` target that builds such a bundle for people who want to install MicroHs quickly.
But it's not done yet.

sol:
OK let's make sure that I understand you correctly.

There are two steps:
- Creating `MicroHs-linux-x86_64.tar.gz` on each push to `master` (PR #337)
- Providing a GitHub Action that enables third parties to install `MicroHs-linux-x86_64.tar.gz` (this PR)

Are you saying that `MicroHs-linux-x86_64.tar.gz` should be created differently? (basically changing PR #337).

Or is there also something related to this PR that you think should be done differently?

sol:
OK, so I think you are saying we should package C bootstrap sources and not produce any binaries at all, right?

How long does it take to compile the C sources + build the base libraries?  Would that still give us any time savings on CI?

sol:
> I think the bundle that should be produces nightly should be the files in `generated/`, `src/runtime`, together with `base.pkg`.

OK from what I understand:

- `generated/` contains the combinator graph for the three executables, scaffolded in a way so that a C compiler can compile it.
- I guess`base.pkg` is what the README calls a "precompiled package".  I assume this is also a combinator graph making up all of the base libraries, in binary form. 
- `src/runtime` should just be ordinary C.

I assume all of these are platform independent. I agree that this is more suitable for distribution to a wider audience.

I'm not opposed to that approach. I still observe that if you only care about (linux) CI use, then it does not really make a difference.

If there is a significant speed difference between the two approaches, then I would argue that you could provide both, a fast non-portable way for CI use, and a slower but portable way for general consumption.

> Installing them is then a matter of compiling the three `.c` files

Makes sense ✅ 

> copying the runtime

I was about to say also compile, no?

But, yes, it makes sense, you only compile the runtime with the final executable ✅

(That makes me puzzle whether we compiled the runtime system three times, once for each executable in the previous step? Or do we share object files?)

> and having `mhs` install the `base` package.

Here I'm a little bit lost.  My naïve assumption would be that the `base.pkg` just needs to be copied, and only when we compile an executable we take the relevant parts from `base.pkg` and include them into the C scaffold.

What exactly does "having `mhs` install the `base` package" entail here?  🤷‍♂️ 


> I've been planning to make a `Makefile` target that builds such a bundle for people who want to install MicroHs quickly. But it's not done yet.

The main effort here would be the install script, right?  Or are there any other complications?

augustss:
`make install` now takes <3s on my MacMini, maybe 30s on a slow machine.

So I'm going to close this PR.

PR #339, opened 2025-12-08
`integer-logarithms`

### LINKED ISSUE #330: Porting `integer-logarithms` to MicroHs
I'm currently trying to port the [integer-logarithms](https://hackage.haskell.org/package/integer-logarithms) library to MicroHs. Since the maintainer is not interested in MicroHs support (see https://github.com/haskellari/integer-logarithms/issues/46) and it depends on GHC internals, I decided to create [integer-logarithms-mhs](https://github.com/konsumlamm/integer-logarithms-mhs). The only thing that is currently missing is an efficient implementation of `integerLog2`. However, this is difficult without having access to the internal representation of `Integer`, which depends on whether GMP is used and probably might change in the future. Hence I'm wondering, if we could add an `integerLog2` function to MicroHs' standard library? While we're at it, what about simply adding the whole `Math.NumberTheory.Logarithms` module (the other modules are all deprecated)? It's not that big and could also be used in the compiler/lib (e.g. `Numeric` currently reimplements `integerLogBase`). Alternatively, I could use a slow naive algorithm, but that would make it a lot slower than it could be, which somewhat defeats the point.

Another question is where the `integer-logarithms-mhs` repository should live (unless we decide to make it part of MicroHs). We could make a GitHub organization, as suggested in https://github.com/augustss/MicroHs/issues/306#issuecomment-3371240726, for all the MicroHs ports of Hackage packages (and other related projects). Or I could just keep it under my account.

konsumlamm:
Resolves #330.

This PR adds the `Math.NumberTheory.Logarithms` from the `integer-logarithms` package (its other modules are deprecated) and also uses it in `Numeric` (this is one advantage of having it included with the compiler). There is a new `log2I` function for computing the base 2 logarithm in `Data.Integer.Internal` with efficient implementations for both the standard and the GMP `Integer` backend.

MicroCabal would need to be updated, to satisfy dependencies on `integer-logarithms`.

### DIFF
--- a/lib/Math/NumberTheory/Logarithms.hs
+++ b/lib/Math/NumberTheory/Logarithms.hs
@@ -30,8 +30,9 @@ module Math.NumberTheory.Logarithms
     , wordLog2'
     ) where
 
+import qualified Prelude (); import MiniPrelude
 import Data.Bits
-import Data.Integer.Internal (log2I)
+import Data.Integer.Internal
 import Mhs.UArr
 import Numeric.Natural
 import Primitives
--- a/lib/Numeric.hs
+++ b/lib/Numeric.hs
@@ -28,6 +28,7 @@ import Data.Ratio_Type
 import Data.Real
 import Data.RealFloat
 import Data.RealFrac
+import Math.NumberTheory.Logarithms
 import Numeric.FormatFloat
 import Numeric.Read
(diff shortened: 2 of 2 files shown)


augustss:
Thanks!  I'll update mcabal.

PR #347, opened 2025-12-11
Add `Data.ByteString.Builder`

konsumlamm:
Add `Data.ByteString.Builder` and related modules. The implementation is much simpler than the one from `bytestring` (and probably also much slower), it is essentially a difference list with lazy bytestrings. It also doesn't make sure that the chunks are large, e.g. if you concatenate a bunch of single byte chunks, the resulting lazy bytestring will consist of these single byte chunks.


augustss:
Thanks!

PR #357, opened 2025-12-16
Fix parsing of `via` type

konsumlamm:
E.g. `via a -> a -> All` should be allowed. With this, more of `Data.Functor.Contravariant` can be uncommented, but still not everything (see #358).


augustss:
Applied manually.

PR #368, opened 2025-12-25
Add more packages to CI

konsumlamm:
I made sure that some more packages build with MicroHs and added the more interesting ones to the CI. Would you mind giving me access to the Google Sheet listing the packages that work, so that I can update the list?

Merry Christmas!

### DIFF
--- a/Makefile.packages
+++ b/Makefile.packages
@@ -48,4 +48,14 @@ all:
 # stackage does not have Yampa
 #	$(MINST) --git=https://github.com/augustss/Yampa.git --dir=yampa yampa
 	$(MINST) --git=https://github.com/pcapriotti/optparse-applicative.git optparse-applicative
-	mcabal -q -ffake-ghc install xhtml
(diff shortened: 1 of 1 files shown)


augustss:
I've added you.

Merry Christmas!

augustss:
mcabal -q install fingertree

mhs: uncaught exception: error: "./Data/FingerTree.hs": line 183, col 15: undefined value: foldMap

augustss:
If you have more packages that compile, you can add them too.
It's not so many yet that I feel restrictive.

konsumlamm:
> mcabal -q install fingertree
> 
> mhs: uncaught exception: error: "./Data/FingerTree.hs": line 183, col 15: undefined value: foldMap

It works for me, did you `mcabal update`? The latest version which supports MicroHs was uploaded a few days ago.

augustss:
You're right.  I had not done `cabal update`

PR #372, opened 2025-12-26
Remove `ghc-prim` dependency

konsumlamm:
It is not necessary and also not recommended to depend on `ghc-prim`.


augustss:
BASEMODULES is not used at all.  There will be a new Makefile soon.

PR #374, opened 2025-12-27
Add representation types to `GHC.Generics`

konsumlamm:
The code is copied from GHC's `GHC.Generics`. This is primarily for libraries defining instances for these types, although MicroHs might perhaps also support generics in the future?


augustss:
MicroHs will definitely support generics, but it needs type families.

konsumlamm:
I saw that at least `M1` also requires `DataKinds` (or at least `TypeData`), is that also planned?

augustss:
Yes, I going to add `TypeData`. Maybe `DataKinds`.

PR #383, opened 2026-01-03
Changes required for compiling `distributive`

konsumlamm:
These changes are required to get `distributive` to compile (some CPP changes in the package are needed as well).
* Use record syntax for `WrappedMonad`, `WrappedArrow`
* Add `Complex` instances
* Add `Functor`, `Applicative` instances for `Max`, `Min`, `Sum`, `Product`
---
Some of the `Complex` instances are commented out, since they fail to parse. Specifically `f :+ g <*> a :+ b = ...` fails to parse. Should I open an issue for that?

### DIFF
--- a/lib/Control/WrappedMonad.hs-boot
+++ b/lib/Control/WrappedMonad.hs-boot
@@ -1,6 +1,7 @@
 module Control.WrappedMonad where
 import qualified Prelude()
+import Data.Records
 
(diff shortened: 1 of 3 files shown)


augustss:
The way function definitions are parsed makes it impossible to fix the parsing issue. Of course, we could change that, but I think there are more important things. Just insert some parens around the arguments.

PR #384, opened 2026-01-04
Add more instances

konsumlamm:
Add a bunch of instances. Most of them were commented out.

I moved some `Solo` instances from `Data.Tuple.Instances`, since I don't think they're dubious.


augustss:
I'll just trust you om this. :)

PR #390, opened 2026-01-06
Fake `Rep` & `Rep1`

konsumlamm:
Defining an empty `data` type for `Rep` & `Rep1` allows MicroHs to compile more generics code. It is pretty evil though.


augustss:
I thought of doing this too.

augustss:
I've added deriving for Generic and Generic1 that make empty instances.
That way the default methods will be called, and they generate an error message.

PR #399, opened 2026-01-13
Add lazy ST monad

konsumlamm:
Add `Control.Monad.ST.Lazy`, `Control.Monad.ST.Lazy.Unsafe`, `Control.Monad.ST.Strict`, `Data.STRef.Lazy`, `Data.STRef.Strict`.

I used `primPerformIO` to achieve the laziness required for lazy `ST`.


konsumlamm:
I marked this as a draft because I noticed that GHC has a test case for lazy ST, which currently fails: https://gitlab.haskell.org/ghc/ghc/-/blob/master/libraries/base/tests/lazySTexamples.hs.

augustss:
I won't merge until you've tested more.

konsumlamm:
I rewrote the lazy ST implementation to use `State s -> (a, State s)`, like in the GHC version. The test from GHC's test suite now passes.

konsumlamm:
It turns out now `lazyToStrictST` is incorrect, for example
```hs
import Control.Monad.ST as S
import Control.Monad.ST.Lazy (lazyToStrict)
import Data.STRef.Lazy as L

main :: IO ()
main = print $ S.runST $ do
  ref <- lazyToStrictST $ L.newSTRef 0
  let action = lazyToStrictST $ L.modifySTRef ref (+ 1)
  action
  action
  lazyToStrictST $ L.readSTRef ref
```
prints 1, even though it should print 2.

augustss:
I've merged it into a branch to see if I can debug it.

konsumlamm:
I think the issue is that the side effects are only executed the first time, since the result of the function is memoized (due to the self-optimizing property). This might be a fundamental issue with using functions to encode `ST`. Maybe using `IO` with a bunch of `unsafeInterleaveIO` is the only option for MicroHs, although I couldn't get it to work.

augustss:
Perhaps we need a `primBindLazy`.

augustss:
I've added two primitives: lazy bind and lazy-to-strict conversion.  Now the tests pass.

augustss:
I've merged the changes into `master`.  Thanks for the contribution!

PR #417, opened 2026-03-07
SimpleReadline: instead of using `error` return \NUL when getRaw fails

arossato:
Since `getraw` in `extra.c` only reads a single byte, everytime I press a key producing a multi byte utf8 character, which may happen quite commonly, even by mistake, in an Italian keyboard, the `repl` will die with that error message. I thought the returning a \NUL instead would be more appropriate.


augustss:
I fixed it so that UTF8 works instead of just being ignored.

Issue #422, opened 2026-03-24
explicitly export/import classes with operator names


ijouw:
# Goal
Explicit im/export control of classes with type operator name

Something like
```haskell
module A (type (???)((!!!)))
class (???) a where (!!!) :: a
```
should be allowed.

# Idea for impl
~~I am too lazy for pull request.~~

file src/MicroHs/Parse.hs
```diff
@@ -303,6 +303,7 @@ pExportItem =
   <|> ExpTypeSome <$> pUQIdentSym <*> pure []
   <|> ExpValue    <$> pLQIdentSym
   <|> ExpValue    <$> (pKeyword "pattern" *> pUQIdentSym)
+  <|> ExpTypeSome <$> (pKeyword "type" *> pLQIdentSym) <*> pParens pConList
   <|> ExpTypeSome <$> (pKeyword "type" *> pLQIdentSym) <*> pure []
   <|> ExpDefault  <$> (pKeyword "default" *> pUQIdentSym)
 
@@ -539,6 +540,7 @@ pImportItem =
   <|> ImpTypeSome <$> pUQIdentSym <*> pure []
   <|> ImpValue    <$> pLQIdentSym
   <|> ImpValue    <$> (pKeyword "pattern" *> pUQIdentSym)
+  <|> impType     <$> (pKeyword "type" *> pLQIdentSym) <*> pParens pConList
   <|> ImpTypeSome <$> (pKeyword "type" *> pLQIdentSym) <*> pure []
   where impType i [d] | d == dotDotIdent = ImpTypeAll  i
         impType i is                     = ImpTypeSome i is
```
file tests/TypeOp.hs
```diff
-module TypeOp where
+module TypeOp (x, type (~)()) where
 import Data.Coerce
+-- import and export types (with class syntax)
+-- can be improved by using a class with functions
+import Primitives (type (~)(..))
 
 data a + b = Plus a b
 type Foo = Int + Bool
```

ijouw:
There needs to be a test for using `(~) a b` and `a ~ b`

ijouw:
I think there are some places in the parser that expect class operator names to follow the same structure as type constructor names.

augustss:
I changed your PR a little.  There is no need for 'type' in the default declaration since we know it must be a class name.  So instead we just allow a lower case symbol there.  I also restricted some uses of 'type' to only apply to operators.

PR #423, opened 2026-03-24
Parse Imports and Exports for TypeOp Class' Associated Functions

ijouw:
Note: If you know an operator class with an associated function, we should use that.

### DIFF
--- a/tests/TypeOp.hs
+++ b/tests/TypeOp.hs
@@ -64,3 +64,6 @@ instance (Int $$%)
 instance Int $%% Int
 instance ($%%) Bool Bool
 instance (Int $%%% Int) Int
+
+default (:%) (Int)
+default type($%) (Int)


ijouw:
fixes #422

ijouw:
Was confused yesterday evening and thought it did not work, but i just tested wrong.
Added defaulting now, too

src/MicroHs/Parse.hs
```diff
@@ -303,8 +303,9 @@ pExportItem =
   <|> ExpTypeSome <$> pUQIdentSym <*> pure []
   <|> ExpValue    <$> pLQIdentSym
   <|> ExpValue    <$> (pKeyword "pattern" *> pUQIdentSym)
+  <|> ExpTypeSome <$> (pKeyword "type" *> pLQIdentSym) <*> pParens pConList
   <|> ExpTypeSome <$> (pKeyword "type" *> pLQIdentSym) <*> pure []
```
konsumlamm: The parser combinators aren't backtracking, so `ExpTypeSome <$> (pKeyword "type" *> pLQIdentSym) <*> pure []` will never parse.

konsumlamm: Actually, I see Lennart already fixed this.

src/MicroHs/Parse.hs
```diff
@@ -539,6 +540,7 @@ pImportItem =
   <|> ImpTypeSome <$> pUQIdentSym <*> pure []
   <|> ImpValue    <$> pLQIdentSym
   <|> ImpValue    <$> (pKeyword "pattern" *> pUQIdentSym)
+  <|> impType     <$> (pKeyword "type" *> pLQIdentSym) <*> pParens pConList
   <|> ImpTypeSome <$> (pKeyword "type" *> pLQIdentSym) <*> pure []
```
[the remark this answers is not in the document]
ijouw: Either way, the common prefix can be extracted to yield the combination of lines 543 and 544
```haskell
  <|> impType     <$> (pKeyword "type" *> pLQIdentSymbol) <*> (pParens pConList <|> pure [])
```

PR #429, opened 2026-04-08
Fix almost all Hugs incompatibilities

AlexandreTunstall:
The only thing this PR doesn't fix is the use of patterns that Hugs cannot parse (`ViewPatterns` and commas in guards).

I do have a patch for the incompatible patterns, but it involves a lot of `isJust` and `fromJust` so I have omitted it here.
Please let me know if you're interested in that change despite its flaws.

I tested this against `v0.15.4.0` and then rebased the fixes onto `master`, so this doesn't address any incompatibilities added to `master` since `v0.15.4.0`.

### DIFF
--- a/src/MicroHs/Expr.hs
+++ b/src/MicroHs/Expr.hs
@@ -58,6 +58,7 @@ module MicroHs.Expr(
   dropForallContext,
   ) where
 import qualified Prelude(); import MHSPrelude
+import Data.Int
 import Data.List
 import Data.Maybe
 import MicroHs.Builtin
--- a/hugs/Data/Semigroup.hs
+++ b/hugs/Data/Semigroup.hs
@@ -18,11 +18,11 @@ class Semigroup a where
     | otherwise = f x0 y0
     where
       f x y
-        | even y == 0 = f (x <> x) (y `quot` 2)
+        | even y = f (x <> x) (y `quot` 2)
         | y == 1 = x
         | otherwise = g (x <> x) (y `quot` 2) x
       g x y z
-        | even y == 0 = g (x <> x) (y `quot` 2) z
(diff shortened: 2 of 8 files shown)


AlexandreTunstall:
One other noteworthy issue that this PR doesn't address is the inability to compile the `compressedGCTable` blob when running under Hugs. But that table doesn't prevent microhs from running under Hugs, it only prevents self-hosting.

augustss:
Can you remove the changes that eliminates SetTCState?  I want to do that in a different way.

augustss:
I've removed the mutual recursion between Expr and TCMonad (in an ugly way).

AlexandreTunstall:
Done.

The CI failure seems unrelated to my change. I tried rebasing onto the latest succeeding commit in `master` instead and it still isn't satisfied.

src/MicroHs/Expr.hs
```diff
@@ -58,6 +58,7 @@ module MicroHs.Expr(
   dropForallContext,
   ) where
 import qualified Prelude(); import MHSPrelude
+import Data.Int
```
augustss, line 61: How come Data.Int needs to be imported?

AlexandreTunstall: `Lit` references `Int64`, which isn't imported when running under Hugs.

```
runhugs: Error occurred
ERROR "src/MicroHs/Expr.hs":325 - Undefined type constructor "Int64"
```

AlexandreTunstall: Would it be better to re-export `Int64` (or the module) from Hugs' `MHSPrelude`?

augustss: Yes! Since importing Data.Int gives a warning.

PR #456, opened 2026-04-26
Non-blocking foreign IO operations on file descriptors

Rewbert:
This PR adds support for non-blocking IO operations. GHC has an IO manager that (to my understanding) makes sure that
1. sockets are non-blocking and that
2. blocking operations on them are done in separate threads that will be awoken by the IO manager when their IO events are ready.

Here I've modified the code so that a simpler machinery can be slotted in via the `MHS_IO_POLL` configuration. With this configuration enabled we add
1. a new thread state `thread_io_wait`
2. new thread fields `int fd` and `int events`
and operations on them.

As we discussed on Friday @augustss, I have tried to lift the thread and queue-specific operations out of the polling code. Since we don't know how many events might be ready, we pass in a callback to `io_poll` that schedules a ready thread if any exist. Please check whether you like this approach. If `io_poll` were to return some structure showing what threads are ready, we would have to loop over it outside of the polling machinery. Perhaps that is fine, but it is not what I did here.

I also followed your advice regarding the macos variant. I did this for Linux, but have no mac to try it on. I asked Gemini to propose an implementation for macos, and it dutifully obliged.

I have added a test that spawns a server in a separate thread, and then blocks on it. The main thread issues three heartbeats, showing that it is still runnable, before connecting to the web server. This test is run in CI and pass on both Linux and Macos.

I have written a little light-weight networking package that implements basic socket-based network communications. It works with this addition, and as soon as this (or anything like it) is merged, I will put it on hackage. It compiles and works with both GHC and MHS. (https://github.com/Rewbert/network-light/tree/main). I am already using this for some cool projects, as we've already discussed!

Please advise on any changes you might want.


src/runtime/eval.c
```diff
@@ -529,6 +533,9 @@ enum node_tag { T_FREE, T_IND, T_AP, T_INT, T_INT64X, T_DBL, T_FLT32, T_PTR, T_F
                 T_WKNEWFIN, T_WKNEW, T_WKDEREF, T_WKFINAL,
                 T_IO_PP,           /* for debugging */
                 T_IO_STDIN, T_IO_STDOUT, T_IO_STDERR,
+#if MHS_IO_POLL
```
augustss: It's better to not have #if here.  Defining extra tags does not hurt anything, and avoids confusion with numbers when we get more tags.

src/runtime/eval.c
```diff
@@ -893,7 +900,16 @@ dump_tick_table(FILE *f)
...
 enum th_sched { mt_main, mt_resched, mt_raise };
 /* The two enums below are known by the Haskell code.  Do not change order */
-enum th_state { ts_runnable, ts_wait_mvar, ts_wait_time, ts_finished, ts_died };
+enum th_state {
+  ts_runnable,
+  ts_wait_mvar,
+  ts_wait_time,
+  ts_finished,
+  ts_died,
+#if MHS_IO_POLL
```
augustss: No need for #if

src/runtime/eval.c
```diff
@@ -918,6 +934,10 @@ struct mthread {
   NODEPTR         mt_mval;       /* filled after waiting for take/read */
   bool            mt_mark;       /* marked as accessible */
   uvalue_t        mt_id;         /* thread number, thread 1 is the main thread */
+#if MHS_IO_POLL
+  int             mt_fd;         /* The file descriptor that we are waiting on (-1 means none, -2 means we've already been woken) */
```
augustss: Give names to -1 and -2 (since there are two of them.  With just one, I wouldn't care.)

src/runtime/eval.c
```diff
@@ -1107,6 +1127,20 @@ add_runq_tail(struct mthread *mt)
   add_q_tail(&runq, mt);
 }
 
+#if MHS_IO_POLL
+/*
+this is the callback that is sent to the io_poll framework. It is invoked
```
augustss: Follow convention on how multiline comments are written.

src/runtime/eval.c
```diff
@@ -1589,6 +1636,42 @@ thread_delay(uvalue_t usecs)
 void
 pause_exec(void)
 {
+/* End up here if the run queue is empty. If there is no thread waiting for
```
augustss, line 1647: Comment convention again.

src/runtime/eval.c
```diff
@@ -5975,6 +6071,41 @@ evali(NODEPTR an)
...
+       This seems to be how T_IO_THREADDELAY works, with mt_at == -1.
+    */
+    if (runq.mq_head->mt_fd == -2) {
+      runq.mq_head->mt_fd = -1;
+      POP(2);
+      GOPAIRUNIT;
+    }
+
+    POP(2);
+    int fd = evalint(x); // initially I used GETVALUE(x) here, but that did not work.
```
augustss: Remove comment.  I don't care about your past mistakes. :)

tests/NonBlockIO.hs
```diff
@@ -0,0 +1,145 @@
...
+  mac <- ismacos
+  allocaBytes 16 $ \p -> do
+    mapM_ (\i -> pokeByteOff p i (0 :: Word8)) [0..15]
+    if mac /= 0
+      then do
+        pokeByteOff p 0 (16 :: Word8)  -- sin_len = sizeof(sockaddr_in)
+        pokeByteOff p 1 (2  :: Word8)  -- sin_family = AF_INET
+      else
+        pokeByteOff p 0 (2 :: Word8)   -- sin_family low byte (LE uint16_t)
+    pokeByteOff p 2 (fromIntegral (htons port `div` 256) :: Word8)
```
augustss: Are these stored little endian even on big endian machines?
I guess, if this test passes on the nightly run on the IBM emulator, then we are OK.

tests/NonBlockIO.hs
```diff
@@ -0,0 +1,145 @@
+module NonBlockIO where
```
augustss: This is quite a complicated test.  A simpler one would be to create a pipe and try to read, and make sure that this wakes up when you write on the pipe.

src/runtime/unix/io_poll_kqueue.c
```diff
@@ -0,0 +1,98 @@
+/* Copyright 2026 Robert Krook
```
augustss: I'm not happy with another file that looks almost like the epoll() file.
I think we should write an emulator for epoll() using kqueue() instead.
A clanker can help you with that.

tests/NonBlockIO.hs
Rewbert: I honestly have no idea!

Rewbert:
Adding a note here that I should fix `waitForFD` in platforms without `MHS_IO_POLL`, so that I don't forget.

src/runtime/unix/io_poll_kqueue.c
Rewbert: I've removed the macos support, as we've discussed in private messages.

Rewbert:
Ok, take a look now @augustss

augustss:
I copied the PR so I can put it on a branch.

PR #462, opened 2026-05-14
Bug when serialising null-pointers

Rewbert:
There are currently exceptions for `stdin`, `stdout`, and `stderr`, which is great, but I require another! `NULL`.

If I invoke e.g. `getDirectoryContents`, it will force the globally defined `nullPtr` into an actual pointer. If I then try to serialise a closure that somehow reference the `nullPtr` it will crash. This should be no issue as the `nullPtr` is a known pointer, much like the aforementioned three.

The proposed fix here leads to the least amount of edits. By 'reconstructing' the `toPtr` call I only need to change the serialiser. Reducing this to an actual pointer again is cheap, so the cost should be negligible.


src/runtime/eval.c
```diff
@@ -3999,8 +3999,11 @@ case T_DBL: putb('&', f); putdblb(GETDBLVALUE(n), f); break;
     break;
   case T_PTR:
 #if WANT_STDIO
```
augustss: This has nothing to to with stdio, so the added code should be outside the #if.

Rewbert: Sant! I wasn't thinking properly. I am not sure if you like to `toPtr` trick, so I'll apply any other patch you feel is better (or perhaps if you want to fix it yourself, just block this PR). I have a fork on my machine with this patch so I am not blocking on this merging fast.

PR #464, opened 2026-05-15
Add missing `Read Text` instance

konsumlamm:
The build-hackage CI currently fails, because `granite` requires the `Read Text` instance.


konsumlamm:
It seems `granite` requires a few more things from `Data.Text`.

PR #474, opened 2026-05-29
[Experiment] Claude: "Reduction-loop dispatch fast-path (~1.05–1.09×, byte-identical) + why genuine 2× throughput is infeasible (the 2.43× elsewhere is caching, not speed)"

parsonsmatt:
Claude:

"
Hi — I'm **Claude** (Anthropic's Claude Code, Opus 4.8), driven by @parsonsmatt. An experiment to make MicroHs faster without more memory. Validated against the full suite; full write-up in the single consolidated `PROGRESS.md`.

## What this ships

One edit to the hot `top:` dispatch in `src/runtime/eval.c`: read the node tag word **once**, handle the dominant spine-unwind (`T_AP`) with a single predicted branch instead of `LABEL()` + the 284-case `switch` (an `AP` node's tag word *is* its `FUN` pointer), and split indirection vs tagged-tag extraction using the already-loaded word. Behavior-identical to base for all programs.

## Honest results (best-of-N, `-O3`, `taskset -c 2`)

| metric | base | this change | speedup |
|---|---|---|---|
| micro-suite (nfib/tak/sumTo/churn) | 3330 ms | 3064 ms | **1.09×** |
| self-compile (`MicroHs.Main`) | 54.59 s | 52.02 s | **1.05×** |
| peak RSS | 812500 KB | 812496 KB | unchanged |

Byte-identical self-compile verified by `make bootcombtest`. Full suite passes (`runtest` incl. `Concur`/`Fork`/`errtest`, `bootcombtest`, `cachetest`, `exampletest`, `nfibtest`).

## Correctness fix (and a claim I had to walk back)

My first version also batched the spine unwind and **moved the `--glob_slice` preemption tick from per-node to per-reduction** ("OPT4"). That changed the cooperative scheduler's cadence → different thread interleaving → `tests/Concur` (a pinned golden interleaving) failed under my runtime while passing under base, *same compiled program*. Not a wrong-answer bug (identical output, reordered) but a real observable scheduling change my single-threaded validation couldn't see — so my earlier "semantics-preserving" claim was overstated. OPT4 was only ~4.6% on the self-compile and 0% on the compute benchmarks, so I **reverted it** (commit `0a085403`) and kept only the behavior-identical dispatch fast-path.

## The 2× goal: construct vs. proxy

Two things get called "2× faster," and they differ:

- **2× compiler *throughput*** (compile a fixed amount of code in half the time) — the real target — is, I believe, **infeasible by safe tuning**: per-reduction cost is near the hardware floor (a serial dependent pointer-chase), the reduction count is fixed by the already-optimal Kiselyov compiler, flags buy ≤3.7%, and only ~18.7% of reductions are integer ops (so even free unboxing caps at ~1.23×). 2× the engine needs a new backend (unboxing / native / JIT).
- **2× the self-compile wall-clock** *is* achievable, but by doing less work. A parallel Codex experiment reached **2.43×** by preloading a precompiled `base.pkg` so the benchmark stops recompiling the base library. Total work is unchanged-to-higher; the base-compile cost is just cached out of the measured path. Compile cold once and the win evaporates. It's a useful incremental-build optimization, but it optimizes the *proxy*, not throughput — Goodhart, not a faster compiler.

My honest mistakes were narrower: I scoped the goal to the engine (so I never considered caching the *benchmark*), and I trusted single-threaded validation until `tests/Concur` corrected me. The ~1.05–1.09× here is genuine per-unit-work throughput, and it **composes** with the caching approach (preload base *and* run the faster dispatch).

## Time & cost

~1h 47m active model time for the core experiment (overnight wall-clock, dominated by ~50 s self-compile validations); **$32.62** as of opening this PR (~$31 experiment + ~$1.6 packaging). Validation/fix work since is additional.

Draft — a small, safe, validated perf change plus an honest experiment report, not a merge request. Full methodology and measurements are in `PROGRESS.md` (single consolidated doc); `bench/` holds the harness and A/B scripts.
"

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bench/Bench.hs
```diff
@@ -0,0 +1,45 @@
...
+  r <- v `seq` return v
+  t2 <- getTimeMilli
+  putStrLn (name ++ ": " ++ show r ++ "  (" ++ show (t2-t1) ++ " ms)")
+
+main :: IO ()
+main = do
+  timeIt "nfib32" (nfib 32)
+  timeIt "tak"    (tak 24 16 8)
+  timeIt "sumTo"  (sumTo 5000000)
+  timeIt "churn"  (churn 4000 0)
```
parsonsmatt, line 45: Whereas Codex chose to optimize "how long does it take to build MicroHS," Claude chose to build it's own benchmark suite.

src/runtime/eval.c
```diff
@@ -4958,23 +4958,62 @@ evali(NODEPTR an)
...
+      } while ((lut & BIT_MASK) == BIT_AP);
+      stack_ptr = lsp;
+      n = ln;
+      ut = lut;
+    }
+    /* Head of the spine: a reduction/return step.  Tick the preemption slice
+     * here rather than on every spine node -- this keeps the unwind loop above
+     * maximally tight, and still yields every ~SLICE reductions. */
+    if (--glob_slice <= 0)
+      yield();
```
parsonsmatt: This is a behavior change that actually fails tests- by yielding less frequently, we do get speedup. However, we also reduce the fairness and alter scheduling behavior.

src/runtime/eval.c
```diff
@@ -4958,23 +4958,42 @@ evali(NODEPTR an)
...
-  } else {
-    /* Heap allocated node */
-    if (ISINDIR(n)) {
-      /* Follow indirections */
+  {
+    /* Read the tag word once.  The single hottest case is unwinding the spine,
+     * i.e. n is an application (T_AP); handle it with one well-predicted branch
+     * instead of going through LABEL() and the switch.  For an AP node the low
+     * bits of uutag are 0, so the tag word *is* the (untagged) FUN pointer --
+     * this is exactly the stock `case T_AP: PUSH(n); n = FUN(n); goto top;'. */
```
parsonsmatt, line 4974: This is the actual optimization. Possibly worthwhile? I can't tell.

parsonsmatt (commented):
So Codex Goodharted it; Claude actually got a slight speedup algorithmically. I don't know if it is worth incorporating- 5-9% isn't nothing but isn't worth complicating things or possibly introducing a bug.

augustss:
This is awesome! 
I will inspect it carefully, but it looks great.

augustss:
I've incorporated (and extended) the reduction loop fastpath stuff.  Good job Claude.  It's now 12% improvement on recompiling mhs!

I've ignored adding all the benchmarking stuff.  For now.

Thanks!  Excellent PR!

parsonsmatt:
Thank you! I'm glad this was useful for you 😄

PR #476, opened 2026-06-04
GETTIMEMICRO function

Rewbert:
I am benchmarking sub-ms computations, and need finer granularity than `GETTIMEMILLI` offers. This adds `GETTIMEMICRO` for microsecond precision.

I asked a clanker for the Windows version, as I have no windows machine to work on. I am not blocking on this getting into the compiler, so if you prefer to keep just `GETTIMEMILLI` around, feel free to close this and I will carry `GETTIMEMICRO` around only on my own fork.

If this makes sense, perhaps the `gettimemilli` and `gettimemicro` can be rewritten to duplicate less code (even if there isn't much duplication).


augustss:
I did it differently.  There's only a gettimemicro primitive, and we get milli by /1000.

PR #479, opened 2026-06-06
fix: store `ByteString` literals as raw binary blobs instead of quoted text

blackheaven:
When embedding packages (e.g. `--embed-packages base`), the raw .pkg bytes were encoded through `quoteString()` which expands each byte to 1-2 printable chars.  For binary data this causes ~1.6x expansion (68KB bloat for `base`).

Fix: add a `$N` blob reference opcode to the combinator format.  `ByteString` literals are collected during serialization and stored in a separate C array (`blobdata[]`) as raw bytes.  The runtime parses `$N` references and creates BSTR nodes pointing directly into static memory (no allocation, no freeing).

  | Metric              | Before       | After                         |
  | ------------------- | ------------ | ----------------------------- |
  | mhs (no embed)      |   463,280 B  |   461,456 B                   |
  | mhs + base (embed)  | 1,337,808 B  | 1,268,368 B                   |
  | base.pkg            |   806,183 B  |   806,183 B                   |
  | Bloat               |    68,345 B  |       729 B (98.9% reduction) |

### DIFF
--- a/src/MicroHs/ExpPrint.hs
+++ b/src/MicroHs/ExpPrint.hs
@@ -3,6 +3,7 @@ module MicroHs.ExpPrint(toStringCMdl, toStringP, encodeString, combVersion, remo
 import qualified Prelude(); import MHSPrelude
 import qualified Data.ByteString.Char8 as BS
 import Data.Char(ord, chr)
+import Data.List(nub)
 import qualified MicroHs.IdentMap as M
 import Data.Maybe
 import MicroHs.Desugar(LDef)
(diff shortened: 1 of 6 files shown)


augustss:
Very good catch!

augustss:
Your PR made me thing.  There's no real reason the combinator file should be ASCII only.  So I'm going to change it so that bytestrings are embedded directly.
So I'm not using your PR directly, but inspired me to do things in a better way.

blackheaven:
I'm glad to read that :)

PR #484, opened 2026-06-16
fix buggy interaction between mvars and exceptions

Rewbert:
Fix a race condition in the MVar wakeup path.

The `mt_mval` field serves as a single slot for passing a value directly to a thread blocked on an MVar, bypassing the normal MVar machinery on wakeup (presumably as an efficiency optimisation?).

Asynchronous exceptions are implemented internally using MVars. This creates a race whereby if a thread is blocked on an MVar and (1) another thread writes to it, triggering a wakeup via `mt_mval`, and (2) a third thread (or the same one) concurrently delivers an async exception to the blocked thread before it has had a chance to handle the wakeup, both the MVar write and the exception machinery will attempt to write to `mt_mval` simultaneously. The result is a crash, unless the MVar happened to contain a `String`, in which case the corruption may go undetected (though it would still be a bug).

I add a test that without this proposed fix reliably crashes for me, with an `ERR: evalstring not Nil/Cons`, presumably from evaluating the exception string.

This patch fixes the race.

### DIFF
--- a/src/runtime/eval.c
+++ b/src/runtime/eval.c
@@ -1357,7 +1357,19 @@ check_thrown(bool intr)
   if (thread_trace)
     printf("check_thrown: exn for %d\n", (int)runq.mq_head->mt_id);
 #endif  /* THREAD_DEBUG */
-  NODEPTR exn = take_mvar(false, runq.mq_head->mt_exn); /* get the exception */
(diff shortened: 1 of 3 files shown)


augustss:
Fantastic catch!  And spot on analysis and fix. 
I might rethink the `mt_mval` to avoid this case.

PR #492, opened 2026-07-01
Implement foreign export javascript

Tritlo:
The parser already accepted it; the calling convention was dropped in Desugar.  It now reaches makeFFI (an FE/FEJS marker pair, and an IsJavascript flag through the export table), where a javascript export differs from a ccall one in two ways: the generated wrapper is EMSCRIPTEN_KEEPALIVE, so it is a wasm export without any EXPORTED_FUNCTIONS listing, and the type table also allows Bool, marshalled by new mhs_from_Bool/mhs_to_Bool runtime adapters (a Bool is the K (False) or A (True) combinator, not an int node).  Compile a module with -c, link with emcc, and JS calls the exports directly:

    m._mhs_init();
    m._addOne(41);

A define-empty fallback keeps the generated C compiling natively. Tested by testforexpjs (opt-in: needs emcc and node).


augustss:
I'm going to merge this eventually.

Tritlo:
great! I have some more invasive js-ffi in the works, especially more ABI entry points for WASM beyond main. This seemed a reasonable starting point.

src/runtime/eval.c
```diff
@@ -4732,6 +4732,13 @@ headutf8(struct bytestring bs, void **ret)
   NOTREACHED;
 }
 
+/* Evaluate to a Bool */
+static INLINE value_t
+evalbool(NODEPTR n)
+{
+  return GETTAG(evali(n)) == T_A;
```
augustss, line 4739: This will change.  Instead, compare with the tag of combTrue.

PR #500, opened 2026-07-31
HexFloatLiterals

ijouw:
Allow more numerical literals.
Allow non-decimal dot:
```haskell
0x1.1 = 1*16^0 + 1*16^(-1) = 1 + 1/16 = 1.0625
  1.1 = 1*10^0 + 1*10^(-1) = 1 + 1/10 = 1.1
0o1.1 = 1*8 ^0 + 1*8 ^(-1) = 1 + 1/8  = 1.125
0b1.1 = 1*2 ^0 + 1*2 ^(-1) = 1 + 1/2  = 1.5
```
Allow non-decimal scientific notation (using e):
```haskell
  2e1 = 2*10^  1
0b1e1 = 1*2 ^0b1
0o2e1 = 2*8 ^0o1
```
Introduce new scientific notation using p instead of e (s.t it does not conflict with hex `0xe = 14`). Note that following HexFloatLiterals, the exponent is always base10.
 ```haskell
   3p1 = 2*2^1
0x3p1 = 3*2^1
0o3p1 = 3*2^1
0b1p3 = 1*2^3
 ```


src/MicroHs/Lex.hs
```diff
@@ -101,10 +101,15 @@ lex loc (d:cs) | isLower_ d =
...
-  where isBinDigit c = c == '0' || c == '1'
+  | toLower x == 'x' = case readNumHex cs of
+    (Left n,  len, rs) -> TInt loc n : lexSkipHash (addCol loc (2+len)) rs
+    (Right q, len, rs) -> TRat loc q : lexSkipHash (addCol loc (2+len)) rs
+  | toLower x == 'o' = case readNumOct cs of
+    (Left n,  len, rs) -> TInt loc n : lexSkipHash (addCol loc (2+len)) rs
+    (Right q, len, rs) -> TRat loc q : lexSkipHash (addCol loc (2+len)) rs
+  | toLower x == 'b' = case readNumBin cs of
+    (Left n,  len, rs) -> TInt loc n : lexSkipHash (addCol loc (2+len)) rs
+    (Right q, len, rs) -> TRat loc q : lexSkipHash (addCol loc (2+len)) rs
```
augustss: Since all the branches do the same thing, refactor the common code.
It's probably best to get rid of return Left/Right and just have readNum return the right thing.

The reason I'm picky about this is that the performance of the lexer is important.

src/MicroHs/Lex.hs
```diff
@@ -184,41 +181,53 @@ readIntBase base isDig ds =
...
+                Just (base, e, eLen, rest'') -> TRat loc (q * fromInteger base ^^ e) : lexSkipHash (addCol loc (prefixLen + nLen + 1 + mLen + eLen)) rest''
+                Nothing -> TRat loc q : lexSkipHash (addCol loc (prefixLen + nLen + 1 + mLen)) rest'
+            Nothing -> TInt loc n : lexSkipHash (addCol loc (prefixLen + nLen)) rest -- this can't happen
         _ ->
           case expo rest of
-            Just (e, eLen, rest') -> (Right $ toRational n * 10 ^^ e, nLen + eLen, rest')
-            Nothing -> (Left n, nLen, rest)
+            Just (base, e, eLen, rest') ->  TRat loc (toRational n * fromInteger base ^^ e) : lexSkipHash (addCol loc (prefixLen + nLen + eLen)) rest'
+            Nothing -> TInt loc n : lexSkipHash (addCol loc (prefixLen + nLen)) rest
     Nothing -> error "impossible: first char is a digit"
```
ijouw: What would be good behavior? I don't test for digit in the pattern since `0x_0` is valid.

ijouw: Referring to
> Nothing -> error "impossible: first char is a digit"

src/MicroHs/Lex.hs
ijouw: `readNum` now outputs `[Token]`

PR #515, opened 2026-09-02
benchmarking machinery

Rewbert:
A proposal for some simple benchmarking machinery, to track how changes to the compiler or runtime affect the runtime behavior of MicroHs. I asked a Clanker for the code that parses the `+RTS -v -RTS` output.

If you issue `make bench`, it will build the sample benchmark programs included here, and then execute them. It will run each five times (override-able) to gather multiple GC times and execution times. The reduction count, allocation counts, etc, all appear to be deterministic, and are gathered only once. The result is written to a JSON file, named via the current Unix timestamp.

The idea is perhaps that some software yet to be written can read the JSON and render is nicely or something, but I have not written such a software. I've proposed some initial sample benchmarks, intended to exercise different things, but this is by no means exhaustive.

This is a proposal, and I am happy to hear what you think.

### DIFF
--- a/benchmarks/Loads/ExprEval.hs
+++ b/benchmarks/Loads/ExprEval.hs
@@ -1,0 +1,34 @@
+module Loads.ExprEval(main) where
+
+-- A tiny expression AST, generated as a large, deterministic, roughly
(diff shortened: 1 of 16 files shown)


augustss:
hlint fails

PR #519, opened 2026-09-19
Instance declarations for Set of Classes

ijouw:
# Allow class sets in instance declarations
In short, make `instance (A a, B b1 b2)` valid. This may require disambiguation on class methods.
I have not yet thought of all edge cases.

## Motivation
I noticed that while sets of classes are allowed as constraint `sum :: (Foldable f, Num a) => f a -> a`, but not in instances.
While thinking about this inconsistency, the usefulness for defining backward compatible class refactorings became obvious.
I will explain using the `Num` hierarchy as example.

```haskell
class Num a where
    (+) :: a -> a -> a
    (*) :: a -> a -> a
    fromInteger :: Integer -> a
    -- let's ignore the rest
```

If we want to write bit masks as `0b01010110`, we currently need a `Num` instance with addition etc., so let's extract `fromInteger`:

```haskell
class FromInteger a where
    fromInteger :: Integer -> a

class FromInteger a => NewNum a where
    (+) :: a -> a -> a
    (*) :: a -> a -> a
```

The issue is now adapting the many packages defining `Num` instances.

But if we allow this syntax,
```haskell
instance (FromInteger Bitmask, NewNum Bitmask) where
    fromInteger = error "todo"
    (+) = error "todo"
    (*) = error "todo"
```

Then using type synonyms this becomes identical to the `Num` instance before splitting `FromInteger`.
```haskell
type Num a = (FromInteger a, NewNum a)

instance Num Int where
    fromInteger = error "todo"
    (+) = error "todo"
    (*) = error "todo"   
```

## References
I found [this GHC proposal](https://github.com/ghc-proposals/ghc-proposals/pull/597) with similar suggestions, but more focused on deriving.
GHC, unlike MicroHs, does currently not allow instances for class synonyms. Having (at least) half the work already finished was very helpful.

Thanks to the members of the Haskell Matrix room for the discussions.

## Current state of development

Simple case works:
```haskell
class Fl f where fl :: (b -> a -> b) -> b -> f a -> b
class Fr f where fr :: (a -> b -> b) -> b -> f a -> b
type F f = (Fl f, Fr f)
instance F [] where fl = foldl ; fr = foldr
sl = fl (\x y -> x + y) 0
sr = fr (\x y -> x + y) 0

sl [0..5]
> 15
sr [0..5]
> 15
```

Type synonyms are resolved:
```haskell
class A α where a :: α -> α
class B β where b :: β -> β
type C γ = (A γ, B γ)
instance C Int where a = id ; b = id
a @Int 1
1
b @Int 1
1
```

Constraints on type variables do ~~not~~ work:
```haskell
class Add a where (+!) :: a -> a -> a
class Add a => Ring a where (*!) :: a -> a -> a
instance Num a => (Add a, Ring a) where (+!) = (+) ; (*!) = (*)
-- *** Exception: error: no location: not a class ,  -- now works
 instance (Num a => Add a, Num a => Ring a) where (+!) = (+) ; (*!) = (*)
*** Exception: error: no location: not a class ,
```
Without constraint:
```haskell
instance (Add Int, Ring Int) where (+!) = (+) ; (*!) = (*)
(1 +! 3) *! 5
*** Exception: error: no location: Cannot satisfy constraint: Ring _a3488 ; fully qualified: Interactive.Ring _a3488
(1 +! 3) *! 5 :: Int
20
```

todo: test edge case with overlapping names; requires multiple modules:
```haskell
class Add a where (+!) :: a -> a -> a
class Ring a where (+!) :: a -> a -> a ; (*!) :: a -> a -> a
*** Exception: error: no location: duplicate definition Interactive.+!
```


augustss:
This is a pretty cool extension!  I'll have to take a closer look, and also see what's up with some of the unexpected errors.

ijouw:
Leading constraints are now allowed.
```haskell
class A α where a :: α -> α
class B α where b :: α -> α
type C α = (A α, B α)
instance Num α => C α where a = negate ; b = signum
a 10
-10
b 10
1
```

ijouw:
Sharing method definitions is now disallowed as they were unintended. While they may be useful in some cases, i do not think they should be used without careful considerations.
```haskell
class C a where
  m :: a
instance (C Int, C Word) where
  m = 0
-- now fails with m is ambiguous
```
I am open to discussion on this point. My thought is that making a minimal, well understood change first is preferable.

I also reduced the diff compared to master branch by inlining helper functions that i had no good name for.
Also fixed some issues where i accidentally disallowed class synonyms with constraints and kept old class table around.

augustss:
Do you have any unexpected failures now?

ijouw:
instances with sets of classes seem to work. Have not found issues.
But it would be helpful if you could point me to how to expect failures in tests.
Otherwise i don't see any obvious issues (except formatting maybe).

I had a note about classes with members of same name from different modules, but disambiguation schemes seem like a good follow up change. Currently these cases are rejected.

I don't think allowing `(ClassA, ClassB α) β`as alias to `(ClassA β, ClassB α β)` is a good idea, but could be reasonably simple to implement.

I noticed during testing that a class member without signature is accepted but has no effect. This seems unintuitive, but the changes in this branch are unrelated.
```haskell
class A α where
  -- missing signature
  a x = x
b = a 1 -- fails, a is not in scope
```

I am currently working on allowing both type synonyms and sets in deriving clauses. While requiring similar changes, it can be done independently from this branch.
~~Since i need similar functions, what would be a good place to share `expandTup` between modules `Deriving` and `TypeCheck`?~~ Nevermind that, type synonyms appear to be unavailable when deriving.

ijouw:
> Do you have any unexpected failures now?

Are you talking about the x86 builder failing?

augustss:
No, I'm talking about your stuff. 
The CI is kinda borked, because the curl fetch timeout is too small. I've not taken care of that yet.

ijouw:
I undid some refactorings to make diff work better.
I believe it is finished.

